PR #2, opened 2019-06-21
Add list_objects API

krisis:


src/minio.rs
```diff
@@ -312,6 +312,50 @@ impl Client {
                     .and_then(|s| xml::parse_bucket_list(s))
             })
     }
+
+    pub fn list_objects(
+        &self,
+        b: &str,
```
dvaldivia: maybe we should start calling the function parameters something more meaningful like `bucket_name` and such

src/minio/xml.rs
```diff
@@ -63,3 +73,88 @@ pub fn get_mk_bucket_body() -> Result<Body, Err> {
...
+        let sizes = node.children().filter(|node| node.has_tag_name("Size"));
+        let storage_classes = node
+            .children()
+            .filter(|node| node.has_tag_name("StorageClass"));
+        for (key, (mtime, (etag, (size, storage_class)))) in
+            keys.zip(mtimes.zip(etags.zip(sizes.zip(storage_classes))))
+        {
+            let sz: i64 = parse_tag_content(size)?;
+            let object_info = ObjectInfo::new(
+                key.text().unwrap(),
```
krisis: Need to add validation for empty tags like Key, modified time etc.

PR #3, opened 2019-06-21
Add listen bucket notification

dvaldivia:
One can subscribe to bucket notifications and get an `impl Stream<Item = NotificationInfo, Error = Err>`

```
c.listen_bucket_notificaion("cattest")
            .map_err(|_e| ())
            .for_each(|x| {
                println!("asdasd {:?}", x);
                Ok(())
            })
```

Also added license header.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "minio-rs"
 version = "0.1.0"
-authors = ["Aditya Manthramurthy <aditya.mmy@gmail.com>", "Krishnan Parthasarathi <krishnan.parthasarathi@gmail.com>"]
+authors = ["MinIO Dev Team <dev@min.io>"]
 edition = "2018"
 
 [dependencies]
@@ -12,6 +12,10 @@ hyper = "0.12.28"
 hyper-tls = "0.3.2"
 ring = "0.14.6"
 roxmltree = "0.6.0"
+serde = "1.0.92"
+serde_derive = "1.0.91"
+serde_json = "1.0.39"
 time = "0.1.42"
+tokio = "0.1.21"
 xml-rs = "0.8.0"
 
--- a/src/minio.rs
+++ b/src/minio.rs
@@ -422,7 +422,8 @@ impl Client {
                         c[0] != SPACE_BYTE[0]
                     })
                     .map(|chunk| {
-                        // Split the chunk by lines and process
+                        // Split the chunk by lines and process.
+                        // TODO: Handle case when partial lines are present in the chunk
                         let chunk_lines = String::from_utf8(chunk.to_vec())
                             .map(|p| {
                                 let lines =
--- a/src/minio.rs
+++ b/src/minio.rs
@@ -216,7 +216,7 @@ impl Client {
         bucket_name: &str,
     ) -> impl Future<Item = Region, Error = Err> {
         let mut qp = Values::new();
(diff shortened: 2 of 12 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "minio-rs"
 version = "0.1.0"
-authors = ["Aditya Manthramurthy <aditya.mmy@gmail.com>", "Krishnan Parthasarathi <krishnan.parthasarathi@gmail.com>"]
+authors = ["Aditya Manthramurthy <aditya.mmy@gmail.com>", "Krishnan Parthasarathi <krishnan.parthasarathi@gmail.com>", "Daniel Valdivia hi@danielvaldivia.com"]
```
donatello:
```suggestion
authors = ["MinIO Dev Team" <dev@min.io>]
```

src/minio.rs
```diff
@@ -308,10 +337,63 @@ impl Client {
                 resp.into_body()
                     .concat2()
                     .map_err(|err| Err::HyperErr(err))
-                    .and_then(move |chunk| b2s(chunk.into_bytes()))
+                    .and_then(move |chunk| chunk_to_string(&chunk))
                     .and_then(|s| xml::parse_bucket_list(s))
             })
     }
+
+    /// get_bucket_location - Get location for the bucket_name.
```
donatello: Update comment.

src/minio.rs
```diff
@@ -308,10 +337,63 @@ impl Client {
...
+            method: Method::GET,
+            bucket: Some(bucket_name.to_string()),
+            object: None,
+            headers: HeaderMap::new(),
+            query: query_params,
+            body: Body::empty(),
+            ts: time::now_utc(),
+        };
+        // We'll be streaming the events out of this
+        let (tx, rx) = mpsc::unbounded_channel();
```
donatello: Why do we need this approach? We can just return a stream from the response future directly (using maybe the flatten_stream() combinator). This approach is spawning an extra future.

src/minio.rs
dvaldivia: With what? :S

src/minio.rs
```diff
@@ -308,10 +335,67 @@ impl Client {
                 resp.into_body()
                     .concat2()
                     .map_err(|err| Err::HyperErr(err))
-                    .and_then(move |chunk| b2s(chunk.into_bytes()))
+                    .and_then(move |chunk| chunk_to_string(&chunk))
                     .and_then(|s| xml::parse_bucket_list(s))
             })
     }
+
+    /// listen_bucket_notificaion - Get bucket notifications for for the bucket_name.
```
donatello: s/for for/for/

src/minio.rs
```diff
@@ -352,10 +381,67 @@ impl Client {
...
+            .map(|resp| {
+                // Read the whole body for bucket location response.
+                resp.into_body()
+                    .map_err(|e| Err::HyperErr(e))
+                    .filter(|c| {
+                        // filter out white spaces sent by the server to indicate it's still alive
+                        c[0] != SPACE_BYTE[0]
+                    })
+                    .map(|chunk| {
+                        // Split the chunk by lines and process
```
donatello:
```suggestion
                        // Split the chunk by lines and process.
                        // TODO: Handle case when partial lines are present in the chunk
```

PR #9, opened 2020-09-24
Add put_object_req method

grumbach:
Hello! I noticed this lib doesn't have a put_object method, here it is :

- added put_object_req function
https://github.com/grumbach/minio-rs/blob/93e5cb1108d9d6c97957c82a0511b728618dd7e9/src/minio.rs#L309

- changed `r` to owned to give `r.body` in `mk_request`
https://github.com/grumbach/minio-rs/blob/93e5cb1108d9d6c97957c82a0511b728618dd7e9/src/minio/api.rs#L43

I'll try to write some docs for this repo when I find some time.


dvaldivia:
thanks for your PR @grumbach !

PR #20, opened 2022-09-04
Add compose_object() API

balamurugana:
Signed-off-by: Bala.FA <bala@minio.io>

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -498,6 +498,106 @@ impl<'a> ClientTest<'_> {
         spawned_task.await;
         assert_eq!(receiver.recv().await.unwrap(), true);
     }
+
+    async fn copy_object(&self) {
+        let src_object_name = rand_object_name();
+
+        let size = 16_usize;
+        self.client
+            .put_object(
+                &mut PutObjectArgs::new(
+                    &self.test_bucket,
+                    &src_object_name,
+                    &mut RandReader::new(size),
+                    Some(size),
+                    None,
+                )
+                .unwrap(),
+            )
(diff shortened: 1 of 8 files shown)


tests/tests.rs
```diff
@@ -498,6 +498,92 @@ impl<'a> ClientTest<'_> {
...
+
+        let object_name = rand_object_name();
+
+        self.client
+            .compose_object(
+                &mut ComposeObjectArgs::new(&self.test_bucket, &object_name, &mut sources).unwrap(),
+            )
+            .await
+            .unwrap();
+
```
harshavardhana, line 576: Can you check head object to ensure the right object size is present on the server ?

tests/tests.rs
```diff
@@ -498,6 +498,92 @@ impl<'a> ClientTest<'_> {
...
+                &CopyObjectArgs::new(
+                    &self.test_bucket,
+                    &object_name,
+                    CopySource::new(&self.test_bucket, &src_object_name).unwrap(),
+                )
+                .unwrap(),
+            )
+            .await
+            .unwrap();
+
```
harshavardhana, line 532: Better to verify the size of the object after copy..

harshavardhana (changes requested):

tests/tests.rs
balamurugana: Done

tests/tests.rs
balamurugana: Done

harshavardhana (approved):

PR #21, opened 2022-09-05
add docs build 'deploy-docs.sh'

harshavardhana:


harshavardhana:
@balamurugana would it be preferred keep this in gh-pages? - I am okay to keep this in master since its easier to update and manage.

balamurugana:
> @balamurugana would it be preferred keep this in gh-pages? - I am okay to keep this in master since its easier to update and manage.

It is better to move to `gh-pages` because
1. `gh-pages` contains generated documentation and `docs` directory will have MinIO style documentation.
2. We need generated documentation only for released versions.
3. Development changes many things which will cause huge diffs in documents ATM.

harshavardhana:
okay @balamurugana let me change it to the `gh-pages` branch.

Issue #27, opened 2023-01-17
Please drop dependency on dead chrono-locale


hlavaatch:
`chrono-locale` crate is dead and does not compile with the latest `chrono`. It is used only in `util.rs` - `to_http_header_value()`.
I suggest dropping the dependency and replacing
```rust
time.formatl("%a, %d %b %Y %H:%M:%S GMT", "C").to_string()
```
with
```rust
time.format("%a, %d %b %Y %H:%M:%S GMT").to_string()
```

Dzordzu:
Any perspectives on merging this fix into master? Or maybe a little help is wanted? @hlavaatch @balamurugana

harshavardhana:
#30 resolves this

PR #28, opened 2023-01-17
Drop dependency on dead chrono-locale crate

hlavaatch:
Solves #27


src/s3/utils.rs
```diff
@@ -72,7 +71,7 @@ pub fn to_amz_date(time: UtcTime) -> String {
 }
 
 pub fn to_http_header_value(time: UtcTime) -> String {
-    time.formatl("%a, %d %b %Y %H:%M:%S GMT", "C").to_string()
+    time.format("%a, %d %b %Y %H:%M:%S GMT").to_string()
```
balamurugana, line 74: What locale is used here?

balamurugana (changes requested):

src/s3/utils.rs
hlavaatch: Looking at the chrono source code, without the non-default "unstable-locales" feature as you use it, it uses a statically defined (in src/format/mod.rs Locales::new(None) ) locale that matches what the normal "C"/"POSIX" locale is.

balamurugana: I would like to get a concrete evidence on that. Here the time must be in `C` or `US-EN` locale irrespective of system's locale. Otherwise it is hard to proceed.

hlavaatch: Let me provide the details:

- (`UtcTime`)[https://github.com/minio/minio-rs/blob/master/src/s3/utils.rs#L32] is a type alias for `chrono::DateTime<Utc>` defined in util.rs
- (`format` method)[https://github.com/chronotope/chrono/blob/main/src/datetime/mod.rs#L762]
calls `DateTime<Tz>::format_with_items`
calls `DelayedFormat::new_with_offset` with offset taken from Utc::Offset creating a DelayedFormat (whioch does not include locale without the "unstable-locales" feature)
ceating a DelayedFormat and returning it
- we call to_string() on the DelayedFormat
- DelayedFormat.to_string() invokes std::fmt::Display::fmt on DelayedFormat
    calls chrono::format::format
    calls format_inner for all the items (with Locale set to None)[https://github.com/chronotope/chrono/blob/main/src/format/mod.rs#L837]
- (`chrono::format::format_inner`)[https://github.com/chronotope/chrono/blob/main/src/format/mod.rs#L503] (creates a Locale using Locales::new\(None))[https://github.com/chronotope/chrono/blob/main/src/format/mod.rs#L511]
- Locales::new(None) without the "unstable-locales" feature (simply returns a statically defined Locales instance with constant arrays that do not depend on system locale in any way and match the standard "C"/"POSIX" locale)[https://github.com/chronotope/chrono/blob/main/src/format/mod.rs#L454]

So, that's what I meant when I said "statically defined" locale.

balamurugana: I would better prefer to construct string what we want irrespective `chrono` internals like below
```rust
format!("{}, {} {} {} GMT", time.weekday(), time.day(), match time.month() {
	1 => "Jan",
	2 => "Feb",
	3 => "Mar",
	4 => "Apr",
	5 => "May",
	6 => "Jun",
	7 => "Jul",
	8 => "Aug",
	9 => "Sep",
	10 => "Oct",
	11 => "Nov",
	12 => "Dec",
	_ => "",
    }, time.format("%Y %H:%M:%S").to_string())
```

harshavardhana:
Not needed after #30

PR #42, opened 2023-08-30
Fix a bug in (un)marshaling ReplicationConfig

HoKim98:


balamurugana (dismissed):

src/s3/types.rs
```diff
@@ -1482,7 +1482,10 @@ pub struct Destination {
 impl Destination {
     pub fn from_xml(element: &Element) -> Result<Destination, Error> {
         Ok(Destination {
-            bucket_arn: get_text(element, "Bucket")?,
+            bucket_arn: element
+                .get_child("Destination")
+                .ok_or_else(|| Error::XmlError("<Destination> tag not found".into()))
+                .and_then(|element| get_text(element, "Bucket"))?,
```
balamurugana, line 1488: You should use current logic in `1628` not here.

balamurugana (changes requested):

src/s3/types.rs
kerryeon: Reverted.

src/s3/types.rs
```diff
@@ -1781,17 +1769,17 @@ impl ReplicationConfig {
         let mut data = String::from("<ReplicationConfiguration>");
 
         if let Some(v) = &self.role {
-            data.push_str("<Status>");
-            data.push_str(&v);
-            data.push_str("</Status>");
+            data.push_str("<Role>");
```
balamurugana, line 1772: Firstly open an issue with complete information about current encode/decode issue of `ReplicationConfig`. I do not understand what you are trying to fix.

balamurugana (changes requested):

HJLebbink:
@HoKim98 @balamurugana Housekeeping here, If you guys think there is still something to fix, please reopen.

PR #49, opened 2023-09-26
Add file-uploader example

donatello:
Also add some basic fields to Cargo.toml to prepare for initial release.


README.md
```diff
@@ -44,17 +49,14 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     // 'asiaphotos-2015.zip' to bucket 'asiatrip'.
     client
         .upload_object(
-            &mut UploadObjectArgs::new(
-                &bucket_name,
-                "asiaphotos-2015.zip",
-                "/home/user/Photos/asiaphotos.zip",
-            )
-            .unwrap(),
+            &mut UploadObjectArgs::new(&bucket_name, "asiaphotos-2015.zip", "/tmp/asiaphotos.zip")
```
balamurugana, line 52: Can we use `"/home/user/Photos/asiaphotos.zip"` instead of `"/tmp/asiaphotos.zip"` like all other SDKs?

examples/file-uploader.rs
balamurugana, line 1: As we cover this in `README.md`, do we still need the same code here?

README.md
donatello: Ok, I had changed it only to test it locally

examples/file-uploader.rs
donatello: Yes, by putting it here we ensure that example is always building. Wanted to make sure the readme example is compiling.

PR #50, opened 2023-09-26
Add builder style constructor for Client

donatello:
Builder style makes it easier to construct the Client because each setting is done via a separate function thus documenting the purpose the setting.

Ref: https://rust-lang.github.io/api-guidelines/type-safety.html#builders-enable-construction-of-complex-values-c-builder

### DIFF
--- a/src/s3/client.rs
+++ b/src/s3/client.rs
@@ -40,6 +40,7 @@ use std::collections::{HashMap, VecDeque};
 use std::fs::File;
 use std::io::prelude::*;
 use std::io::Read;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use xmltree::Element;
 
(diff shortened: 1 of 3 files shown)


src/s3/client.rs
```diff
@@ -203,6 +204,91 @@ fn parse_list_objects_common_prefixes(
...
+        self
+    }
+
+    /// Set file for loading a trust certificate.
+    pub fn ssl_cert_file(mut self, ssl_cert_file: Option<&Path>) -> Self {
+        self.ssl_cert_file = ssl_cert_file.map(PathBuf::from);
+        self
+    }
+
+    /// Set the ignore_cert_check.
```
balamurugana, line 247: You could have `/// Set flag to ignore certificate check`

src/s3/client.rs
```diff
@@ -203,6 +204,91 @@ fn parse_list_objects_common_prefixes(
...
+        self.ignore_cert_check = ignore_cert_check;
+        self
+    }
+
+    /// Build the Client.
+    pub fn build(self) -> Result<Client, Error> {
+        let mut builder = reqwest::Client::builder().no_gzip();
+
+        if let Some(v) = self.user_agent {
+            builder = builder.user_agent(v);
```
balamurugana, line 258: In user_agent, SDK specific information must be present. Users are allowed to add additional app_name/app_version like https://github.com/minio/minio-py/blob/master/minio/api.py#L493. We should not allow to override entire value.

src/s3/client.rs
donatello: Done

src/s3/client.rs
donatello: Done

balamurugana (approved):

PR #51, opened 2023-09-26
Add FromStr instance to BaseURL

donatello:


README.md
```diff
@@ -13,7 +13,7 @@ use minio::s3::http::BaseUrl;
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
-    let base_url = BaseUrl::from_string("https://play.min.io".to_string()).unwrap();
+    let base_url = "https://play.min.io".parse::<BaseUrl>()?;
```
balamurugana, line 16: Can we embed `./examples/file-uploader.rs` here? If not possible, we could give a link to `file-uploader.rs` instead of duplicating content of the file here.  We could easily forget to update both the content.

src/s3/http.rs
```diff
@@ -379,80 +481,8 @@ impl BaseUrl {
     /// // Get base URL from IPv6 address
     /// let base_url = BaseUrl::from_string("[0:0:0:0:0:ffff:c0a8:7c3f]:9000".to_string()).unwrap();
     /// ```
+    #[deprecated(since = "0.1.0", note = "please use `from_str` instead")]
```
balamurugana, line 484: We could remove this warning and remove `from_string()` as we haven't made any release yet.

README.md
donatello: Removed it. There is no way to embed a live version of this code in the readme afaik.

src/s3/http.rs
donatello: Done.

balamurugana (approved):

balamurugana (approved):

PR #52, opened 2023-09-27
fix: Set user agent always

donatello:


balamurugana (commented):
You could also fix https://github.com/minio/minio-rs/blob/master/src/s3/client.rs#L345

donatello:
> You could also fix https://github.com/minio/minio-rs/blob/master/src/s3/client.rs#L345

Oh didn't see that - fixed now.

balamurugana (approved):

PR #53, opened 2023-09-28
Add doc for list_objects

donatello:


src/s3/args.rs
```diff
@@ -1116,21 +1116,33 @@ pub struct ListObjectsArgs<'a> {
     pub extra_headers: Option<&'a Multimap>,
     pub extra_query_params: Option<&'a Multimap>,
     pub region: Option<&'a str>,
+    /// Bucket to list
```
balamurugana, line 1119: This is obvious argument to to list objects API. Either remove the comment or add appropriately like `To fetch list of objects on this bucket`

src/s3/args.rs
```diff
@@ -1116,21 +1116,33 @@ pub struct ListObjectsArgs<'a> {
...
+    /// Used only with GetObjectVersions.
+    pub version_id_marker: Option<&'a str>,
+    /// MinIO extension for ListObjectsV2.
+    pub include_user_metadata: bool,
     pub recursive: bool,
+    /// Set this to use ListObjectsV1. Defaults to false.
     pub use_api_v1: bool,
+    /// Set this to include versions.
     pub include_versions: bool,
+    /// A function to process results from each listing request.
```
balamurugana, line 1145: I think `Callback function to get results of object listing` is more appropriate here.

src/s3/client.rs
```diff
@@ -2750,6 +2750,11 @@ impl Client {
         })
     }
 
+    /// List objects. This function retrieves objects and optionally versions
+    /// too and also supports ListObjectsV1 depending on the argument passed. It
+    /// performs requests successively and returns results to the provided
+    /// `results_fn`. Processing can be stopped by returning false in the
+    /// `results_fn`.
```
balamurugana, line 2757: This comment talks too much internals. I think
```
List objects with version information optionally. `results_fn` callback function is repeatedly called with object information and returning false from the callback stops further listing.
```

balamurugana (approved):

PR #54, opened 2023-09-28
Add streaming functionality for list objects

donatello:
Currently, there is a problem in `client.list_objects` - we cannot access the prefixes within a path, only the objects (when not traversing recursively). This is because the `result_fn` only takes object properties:

https://github.com/minio/minio-rs/blob/8ecabea0d21ad437e6c39cb5020ab27f2425edea/src/s3/args.rs#L1115-L1135

Results function is called as follows:

https://github.com/minio/minio-rs/blob/8ecabea0d21ad437e6c39cb5020ab27f2425edea/src/s3/client.rs#L2810-L2835

So looks like no way to iterate over prefixes in this function interface.

Of course we can use the lower level `list_objects_v2` in the client to do this, but maybe we can make a different interface that is easier to use.

This PR is currently in draft for discussion.


src/s3/args.rs
```diff
@@ -1048,12 +1048,10 @@ impl<'a> ListObjectsV2Args<'a> {
...
     /// ```
     /// use minio::s3::args::*;
-    /// let args = ListObjectsV2Args::new("my-bucket").unwrap();
+    /// let args = ListObjectsV2Args::new("my-bucket");
     /// ```
-    pub fn new(bucket_name: &'a str) -> Result<ListObjectsV2Args<'a>, Error> {
-        check_bucket_name(bucket_name, true)?;
-
-        Ok(ListObjectsV2Args {
+    pub fn new(bucket_name: &'a str) -> ListObjectsV2Args<'a> {
```
donatello, line 1053: I am thinking we do this bucket validity check commonly in `client.execute` so that all these constructors do not have to return `Result<_, Error>`

src/s3/client/paginated.rs
```diff
@@ -0,0 +1,96 @@
...
+    bucket: String,
+    delimiter: Option<String>,
+    encoding_type: Option<String>,
+    max_keys: Option<u16>,
+    prefix: Option<String>,
+    start_after: Option<String>,
+    continuation_token: Option<String>,
+    fetch_owner: bool,
+    include_user_metadata: bool,
+}
```
donatello, line 23: We can avoid this extra struct by removing the lifetime parameter from `ListObjectsV2Args<'a>`. Because of references in the struct, it becomes complicated to make a stream out of it.

tests/tests.rs
```diff
@@ -454,6 +455,66 @@ impl ClientTest {
...
+                .unwrap();
+            names.push(object_name);
+        }
+
+        let mut list_stream = self
+            .client
+            .list_objects_v2_stream(&ListObjectsV2Args::new(&self.test_bucket))
+            .await;
+
+        while let Some(list_resp) = list_stream.next().await {
```
donatello, line 490: Each `list_resp` has type `ListObjectsV2Response` so this is easy to use.

donatello: We can easily add similar streaming interface for other listing APIs too.

src/s3/args.rs
balamurugana: That's not a good idea. We should stop on invalid values right away to avoid further processing. There are many other parameters in different APIs to validate arguments in `new()`. Having different behavior of different APIs will look odd.

src/s3/client/paginated.rs
balamurugana: Apart from higher level listing, we also provide lower level S3 APIs to be used by advanced users. As long as it doesn't break the user experience, I am good.

I am fine by having streaming support as a addition to current callback functionality. At the time of writing I evaluated both the approaches and chose callback approach because of the nature of List Objects S3 APIs. As we have result in hand as vector, the vector can be streamed. If someone wants streaming, they could write it on top level this vector.

balamurugana:
> So looks like no way to iterate over prefixes in this function interface.
> 
> Of course we can use the lower level `list_objects_v2` in the client to do this, but maybe we can make a different interface that is easier to use.

@donatello You have to use lower level S3 APIs for this specific needs. The objective of higher level `list_objects()` is to provide one view for all lower level object listing APIs. As you are interested only with prefixes and there is no way to fetch prefixes without object information as per S3 specification, using lower level S3 APIs is the right choice.

donatello:
> > So looks like no way to iterate over prefixes in this function interface.
> > 
> > Of course we can use the lower level `list_objects_v2` in the client to do this, but maybe we can make a different interface that is easier to use.
> 
> @donatello You have to use lower level S3 APIs for this specific needs. The objective of higher level `list_objects()` is to provide one view for all lower level object listing APIs. As you are interested only with prefixes and there is no way to fetch prefixes without object information as per S3 specification, using lower level S3 APIs is the right choice.
> 

@balamurugana I want to actually list folders and objects at a prefix and get all of them. Adding a streaming API for this should be ok. We already have it in other SDKs.

donatello:
@balamurugana This is ready.

src/s3/args.rs
```diff
@@ -1026,22 +1026,22 @@ impl<'a> ListObjectsV1Args<'a> {
...
 
 /// Argument for [list_objects_v2()](crate::s3::client::Client::list_objects_v2) S3 API
-pub struct ListObjectsV2Args<'a> {
-    pub extra_headers: Option<&'a Multimap>,
-    pub extra_query_params: Option<&'a Multimap>,
-    pub region: Option<&'a str>,
-    pub bucket: &'a str,
-    pub delimiter: Option<&'a str>,
-    pub encoding_type: Option<&'a str>,
+pub struct ListObjectsV2Args {
```
balamurugana, line 1029: Can you modify `list_objects()` API to support streaming generically instead of only to ListObjectsV2 API?

donatello: Done, PTAL @balamurugana

src/s3/args.rs
```diff
@@ -1112,70 +1206,64 @@ impl<'a> ListObjectVersionsArgs<'a> {
...
 
         Ok(ListObjectsArgs {
             extra_headers: None,
             extra_query_params: None,
             region: None,
-            bucket: bucket_name,
+            bucket: bucket_name.to_owned(),
             delimiter: None,
-            use_url_encoding_type: true,
+            encoding_type: EncodingType::default(),
```
balamurugana, line 1266: There is no need to have type for encoding. For any `EncodingType`, we handle only one kind which is `url` encoding i.e. if encoding type is `url`, we return decoded values else we return as is.

Having `use_url_encoding_type bool` is good enough for all cases.

src/s3/types.rs
```diff
@@ -26,7 +26,7 @@ use xmltree::Element;
 
 #[derive(Clone, Debug, Default)]
 /// Contains information of an item of [list_objects()](crate::s3::client::Client::list_objects) API
-pub struct Item {
+pub struct ListItem {
```
balamurugana, line 29: `Item` is used generically across all SDKs. `ListItem` looks odd.

tests/tests.rs
```diff
@@ -421,19 +421,78 @@ impl ClientTest {
...
             .await
             .unwrap();
 
+        self.client
+            .remove_bucket(&RemoveBucketArgs::new(&bucket_name).unwrap())
+            .await
+            .unwrap();
+    }
+
+    async fn list_objects_v2_stream(&self) {
```
balamurugana, line 457: This test is no more required

src/s3/args.rs
donatello: With `use_url_encoding_type` we are making it opt-in to use the encoding parameter. I wanted to make it the default because the purpose of the parameter is to handle encoding unicode characters in object names that cannot be represented in the xml response - quoted from https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html


> [encoding-type](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html#API_ListObjects_RequestSyntax)
> 
> Requests Amazon S3 to encode the object keys in the response and specifies the encoding method to use. An object key can contain any Unicode character; however, the XML 1.0 parser cannot parse some characters, such as characters with an ASCII value from 0 to 10. For characters that are not supported in XML 1.0, you can add this parameter to request that Amazon S3 encode the keys in the response.
> 
> Valid Values: url
> 

This is a good reason to always set `encoding-type=url` (i.e. default setting) because user will not run into this encoding issue mentioned.

I used a type for it to allow for future changes - in case a new encoding type is added, it would be easy to fix the library with a minor update (instead of a breaking change). If a new encoding type `foo` is added in the future, anyone can start using it as just `EncodingType::Other("foo")`, and we can update the library adding `EncodingType::Foo`. With a bool we just lose all these options.

If you think it's overkill I'll change it to `use_url_encoding_type` and set it to `true` by default.

src/s3/types.rs
donatello: `Item` is too generic here - it clashes with other common types in rust like for example https://docs.rs/futures-core/0.3.28/futures_core/stream/trait.Stream.html and https://doc.rust-lang.org/nightly/std/iter/index.html#iterator

`ListItem` looks fine in the function signature doc (I think):
image

Most users would interact with only one of our SDKs and that usage/experience should be good - almost no one will notice differences in the SDKs.

tests/tests.rs
donatello: Will remove this.

src/s3/types.rs
donatello: Maybe `ListedItem` is better?

src/s3/args.rs
balamurugana: `EncodingType` is useful if we handle more than one encoding type. Practically it is either `url` type or none. `use_url_encoding_type` tells what `minio-rs` does with object list xml response. We could keep the current behavior i.e. `use_url_encoding_type = true` as default.

src/s3/types.rs
balamurugana: I see the name conflict. May be like minio-go's `ObjectInfo` or just `ListEntry`?

src/s3/args.rs
donatello: Done @balamurugana

src/s3/types.rs
donatello: Done, renamed to `ListEntry`

src/s3/client/list_objects.rs
```diff
@@ -0,0 +1,561 @@
...
+                            Some((Ok(resp), (client, args)))
+                        }
+                    }
+                    Err(e) => Some((Err(e), (client, args))),
+                }
+            },
+        ))
+    }
+
+    pub async fn list_objects_v2_stream(
```
balamurugana, line 457: This function can be private. Users should use `list_objects()` for streaming.

balamurugana (approved):

PR #56, opened 2023-10-06
Add builder style for list_objects

donatello:


donatello:
@balamurugana this is ready

src/s3/builders/list_objects.rs
```diff
@@ -0,0 +1,737 @@
...
+    max_keys: Option<u16>,
+    prefix: Option<String>,
+    marker: Option<String>,
+}
+
+#[async_trait]
+impl ToStream for ListObjectsV1 {
+    type Item = ListObjectsV1Response;
+
+    async fn to_stream(self) -> Box<dyn Stream<Item = Result<Self::Item, Error>> + Unpin + Send> {
```
balamurugana, line 72: This is not a purpose of low level S3 method. We would need to return single parsed response for single request. This defeats the purpose i.e. custom list implementation is not possible.

src/s3/builders/list_objects.rs
```diff
@@ -0,0 +1,737 @@
...
+        if self.use_api_v1 {
+            let stream = ListObjectsV1::from(self).to_stream().await;
+            Box::new(stream.map(|v| v.map(|v| v.into())))
+        } else if self.include_versions {
+            let stream = ListObjectVersions::from(self).to_stream().await;
+            Box::new(stream.map(|v| v.map(|v| v.into())))
+        } else {
+            let stream = ListObjectsV2::from(self).to_stream().await;
+            Box::new(stream.map(|v| v.map(|v| v.into())))
+        }
```
donatello, line 622: The `ToStream` instance for the lower level types is used in the `ToStream` instance for higher level `ListObjects` call.

src/s3/builders/list_objects.rs
donatello: @balamurugana Please see the other comments - the `ToStream` instance is needed for the higher level call's implementation. For custom implementation one can use `S3Api::send()` for each other the lower level APIs.

donatello:
@balamurugana PTAL

PR #57, opened 2023-10-06
Add builder for ListenBucketNotification

donatello:
We can take this after https://github.com/minio/minio-rs/pull/56


donatello:
This is also ready @balamurugana

PR #60, opened 2023-11-02
Remove dynamically dispatched fields

juanscr:
Dynamic dispatch is normally a good option to speed up development on high-level code, specially projects focusing more on an outside client consuming it without any interaction with rust internals. On the other hand, using it in a library that should have the least overhead as possible is not ideal. Additionally,  some of the fields that implemented dynamic dispatch where actually hurting implementations in any multi-threaded framework such as `tokio` (`axum`, etc.) as the dynamic fields didn't implemented send so they can be used safely between threads.

In this pull request, we just rewrote many of the dynamic dispatched fields to generic arguments to avoid any weird overhead and improve integrations with other frameworks.

## Important thing to notice
By changing it to generics, we noticed that the instantiation of many of the objects turned a bit ugly as it needed to specify the server side encryption type as it is not passed in the constructor of many of the structs. Maybe consider setting a fixed `Sse` object to avoid this? The changes presented here are fine but maybe it is better to improve the usability by not having to specify that trait.

### DIFF
--- a/examples/file-uploader.rs
+++ b/examples/file-uploader.rs
@@ -2,6 +2,7 @@ use minio::s3::args::{BucketExistsArgs, MakeBucketArgs, UploadObjectArgs};
 use minio::s3::client::Client;
 use minio::s3::creds::StaticProvider;
 use minio::s3::http::BaseUrl;
+use minio::s3::sse::SseCustomerKey;
 
 #[tokio::main]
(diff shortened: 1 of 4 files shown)


xunafay (approved):
Works in my tokio runtime codebase

functorism:
What's preventing this from being merged? I'm now using https://github.com/1ntervalica/minio-rs/tree/master and I can confirm that it indeed solves the issue described in https://github.com/minio/minio-rs/issues/66

PR #92, opened 2024-06-17
allow optional disk sync for download_object()

HowHsu:
Add an argument to allow not syncing the file to disk after get_object_old()  and leave it to the operating system's cache strategy. This is useful in performance sensitive scenarios.


HowHsu:
Btw, this breaks API compatibility

HJLebbink:
@HowHsu Sorry for the delay in getting back to you. I've added this option in the new API, and it should be available soon!

HJLebbink:
@HowHsu It looks like `download_object` has been deprecated and removed. I will put together a handy section with some commonly used functions like `download_objects`, `purge_buckets`, etc.

PR #100, opened 2024-08-29
Respect content_type prop

vajexal:
For now `PutObjectContent.content_type` doesn't work, this pr fixes it

### DIFF
--- a/src/s3/builders/put_object.rs
+++ b/src/s3/builders/put_object.rs
@@ -52,6 +52,7 @@ pub struct CreateMultipartUpload {
     tags: Option<HashMap<String, String>>,
     retention: Option<Retention>,
     legal_hold: bool,
+    content_type: Option<String>,
 }
 
 impl CreateMultipartUpload {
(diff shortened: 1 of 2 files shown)


src/s3/builders/put_object.rs
```diff
@@ -52,6 +52,7 @@ pub struct CreateMultipartUpload {
     tags: Option<HashMap<String, String>>,
     retention: Option<Retention>,
     legal_hold: bool,
+    content_type: Option<String>,
```
donatello, line 55: We need to add method to be able to set this parameter in the CreateMultipartUpload struct as the struct members are not public

donatello (changes requested):
Thanks for the PR! It needs one minor change to be complete.

src/s3/builders/put_object.rs
vajexal: done

donatello (approved):

PR #102, opened 2024-10-25
feat: allow multiple trust certs in cert file

donatello:
The `.ssl_cert_file()` option now can read files with multiple certificate to trust. This is useful when using a single client instance to access many minio servers.


HJLebbink (approved):
LGTM. Works as advertised.

HJLebbink (approved):
LGTM, but it seems you reverted almost all your changes?

donatello:
> LGTM, but it seems you reverted almost all your changes?

No this looks right. We are adding all the certs from the file now.

PR #109, opened 2025-01-09
Added Object Prompt API Support

HJLebbink:


Cargo.toml
```diff
@@ -61,4 +61,5 @@ clap = { version = "4.5.23", features = ["derive"] }
 quickcheck = "1.0.3"
 
 [[example]]
-name = "file-uploader"
+#name = "file_uploader"
+name = "object_prompt"
```
donatello, line 65: This probably removes the file_uploader example? maybe this needs one example section per example program in the crate.

HJLebbink:
@donatello The example I added uploads a picture to play.min.io and runs an object prompt with the question, "What is it about?" However, it seems play.min.io doesn't support object prompts yet. To run the example, you'll need to use a more up-to-date object store.

PR #110, opened 2025-01-14
Refactor get bucket encryption

HJLebbink:
Hi @donatello,

This PR refactors (as a trial) `get_bucket_encryption` in a similar way to the already refactored `get_bucket_versioning`. What’s bothering me, though, is that I can’t seem to reuse `BucketCommon`, which I assume was created to serve as a shared struct for both `get_bucket_encryption` and `get_bucket_versioning`.

No matter what I try, I can’t get it to compile using the obvious implementation:

```
pub type GetBucketEncryption = BucketCommon;
pub type GetBucketVersioning = BucketCommon;
```
For now, I’ve copy-pasted it into a `BucketCommon2` to make it work. Do you have any suggestions for a clean solution?


donatello:
What i think we can try to use a phantom data type - https://doc.rust-lang.org/std/marker/struct.PhantomData.html

The idea is something like the code below. `BucketCommon` becomes generic and we use set type alias like:

```rust 
pub type GetBucketVersioning = BucketCommon<GetBucketVersioningPhantomData>;

pub type GetBucketEncryption = BucketCommon<GetBucketEncryptionPhantomData>;
``` 

where the `*PhantomData` types are empty structs. More complete example below. In my (somewhat basic) testing, this idea looks like it could work. The `new` and trait methods are tied to the type aliases and not the generic `BucketCommon<A>` type. 

```rust
#[derive(Clone, Debug)]
pub struct BucketCommon<A> {
    client: Option<Client>,

    extra_headers: Option<Multimap>,
    extra_query_params: Option<Multimap>,
    region: Option<String>,
    bucket: String,

    operation: PhantomData<A>,
}

impl<A> BucketCommon<A> {
...
}

pub struct GetBucketVersioningPhantomData;

pub type GetBucketVersioning = BucketCommon<GetBucketVersioningPhantomData>;

impl GetBucketVersioning {
    pub fn new(bucket_name: &str) -> Self {
        Self {
            client: None,
            extra_headers: None,
            extra_query_params: None,
            region: None,
            bucket: bucket_name.to_owned(),
            operation: PhantomData,
        }
    }
}

impl S3Api for GetBucketVersioning {
    type S3Response = GetBucketVersioningResponse;
}

impl ToS3Request for GetBucketVersioning {
...
}

pub struct GetBucketEncryptionPhantomData;

pub type GetBucketEncryption = BucketCommon<GetBucketEncryptionPhantomData>;

impl GetBucketEncryption {
    pub fn new(bucket_name: &str) -> Self {
        Self {
            client: None,
            extra_headers: None,
            extra_query_params: None,
            region: None,
            bucket: bucket_name.to_owned(),
            operation: PhantomData,
        }
    }
}

impl S3Api for GetBucketEncryption {
    type S3Response = GetBucketEncResponse;
}
```

HJLebbink:
@donatello Can you please take a look

donatello:
> @donatello Can you please take a look

Apologies for my delay. Checking now!

src/s3/builders/bucket_common.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub(crate) extra_query_params: Option<Multimap>,
+    pub(crate) region: Option<String>,
+    pub(crate) bucket: String,
+
+    _operation: PhantomData<A>,
+}
+
+impl<A: Default> BucketCommon<A> {
+    pub fn new(bucket: &str) -> Result<BucketCommon<A>, Error> {
+        check_bucket_name(bucket, true)?;
```
donatello, line 38: We can move this check into the individual `to_s3request` trait methods. This lets us avoid returning an error here and leads to an API that is smoother to use in the sense that there is only one place where all errors are returned (in the send() call):

```rust
let r = client.get_bucket_encryption("mybucket").send().await;
// Now handle error value of `r`
```

instead of 
```rust
let r1 = client.get_bucket_encryption("mybuclet");
let r2 = r1.unwrap() // or some actual error handling
let r3 = r2.send().await.unwrap() // or more actual error handling
```

src/s3/client/get_bucket_versioning.rs
```diff
@@ -0,0 +1,27 @@
...
+//! S3 APIs for bucket objects.
+
+use super::Client;
+use crate::s3::builders::GetBucketVersioning;
+use crate::s3::error::Error;
+
+impl Client {
+    /// Create a GetBucketVersioning request builder.
+    pub fn get_bucket_versioning(&self, bucket: &str) -> Result<GetBucketVersioning, Error> {
+        Ok(GetBucketVersioning::new(bucket)?.client(self))
```
donatello, line 25: Similarly we can avoid returning an error here.

donatello (changes requested):
Other than the error handling change suggested, this LGTM!

src/s3/client/get_bucket_encryption.rs
```diff
@@ -18,10 +18,12 @@
 use super::Client;
 use crate::s3::builders::GetBucketEncryption;
 use crate::s3::error::Error;
+use crate::s3::utils::check_bucket_name;
 
 impl Client {
     /// Create a GetBucketEncryption request builder.
     pub fn get_bucket_encryption(&self, bucket: &str) -> Result<GetBucketEncryption, Error> {
-        Ok(GetBucketEncryption::new(bucket)?.client(self))
+        check_bucket_name(bucket, true)?;
```
donatello, line 26: @HJLebbink I mean to move it into the `to_s3request()` method for the GetBucketEncryption type - it is similar to how its done for say get object here https://github.com/minio/minio-rs/blob/master/src/s3/builders/get_object.rs#L174

HJLebbink: Yeah your are right, wan't paying attention.

donatello (approved):

PR #112, opened 2025-01-30
Update README.md - dead links

andrenatal:
Fixing links in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,8 @@ Run the examples from the command line with:
 
 ### file_uploader.rs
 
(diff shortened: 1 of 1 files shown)


klauspost:
What about the [MinIO Rust Client API Reference](https://minio-rs.min.io/) - link - can't we also fix that - or add it here?

donatello:
> What about the [MinIO Rust Client API Reference](https://minio-rs.min.io/) - link - can't we also fix that - or add it here?

We need to take that up separately. This library is going through lots of interface changes right now.

donatello (approved):

PR #114, opened 2025-02-03
Refactor set_bucket_versioning

HJLebbink:
Refactored set_bucket_versioning. 

1 down, many to go...


HJLebbink:
@donatello Could you please take a look

src/s3/builders/set_bucket_versioning.rs
```diff
@@ -0,0 +1,128 @@
...
+        self.extra_query_params = extra_query_params;
+        self
+    }
+
+    pub fn region(mut self, region: Option<String>) -> Self {
+        self.region = region;
+        self
+    }
+
+    pub fn status(mut self, status: bool) -> Self {
```
donatello, line 66: Instead of a "status" boolean, I would prefer making it clearer with an enum:
```rust
enum VersioningStatus {
    Enabled
    Suspended
}

pub fn versioning_status(mut self, status: VersioningStatus) -> Self {
```

This is to avoid terminology confusion - we cannot rollback versioning once it is enabled, we can only suspend it (upload of new versions will overwrite the latest version object, but older versions will be preserved).

donatello (changes requested):

src/s3/builders/set_bucket_versioning.rs
HJLebbink: I now have:

```rust
pub enum VersioningStatus {
    /// **Enable** object versioning in given bucket.
    Enabled,
    /// **Suspend** object versioning in given bucket.
    #[default]
    Suspended,
}
````

@donatello, I still have a question about the default behavior. When a bucket is created, it doesn't have a versioning status: it’s neither enabled nor suspended. In this ‘None’ state, would you still want it to return VersioningStatus as Suspended?

Comment (HJLebbink) [reply src/s3/builders/set_bucket_versioning.rs:66]: Instead of wrapping the VersioningStatus in a Option, what might be more intuitive is:
```rust
pub enum VersioningStatus {
    /// **Enable** object versioning in given bucket.
    Enabled,
    /// **Suspend** object versioning in given bucket.
    Suspended,
    #[default]
    None,
}
```

donatello (approved):
LGTM

PR #118, opened 2025-02-13
refactored tests into 24 separate tests

HJLebbink:
Note: depends on https://github.com/minio/minio-rs/pull/114

Refactored one large test into separate tests. It runs 2 twice as fast as a result

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ xmltree = "0.11.0"
 
 [dev-dependencies]
 async-std = { version = "1.13.0", features = ["attributes", "tokio1"] }
-clap = { version = "4.5.23", features = ["derive"] }
+clap = { version = "4.5.27", features = ["derive"] }
 quickcheck = "1.0.3"
 
 [[example]]
--- a/src/s3/builders/put_object.rs
+++ b/src/s3/builders/put_object.rs
@@ -35,7 +35,7 @@ use crate::s3::{
 use super::{ObjectContent, SegmentedBytes};
 
 /// Argument for
-/// [create_multipart_upload()](crate::s3::client::Client::create_multipart_upload)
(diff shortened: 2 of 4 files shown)


donatello:
Please fix conflicts @HJLebbink

HJLebbink:
> Please fix conflicts @HJLebbink

@donatello I rebased. Two minor issues with the code. 

  1.  Every test is responsible for removing the stuff it created. But if the test fails the `remove_bucket_helper` may not be called.  
  2.  TestContext.new_from_env (that reads the environment variables) only works with the configuration used by the build server. For a setup with play.min.io or my local box, it has bugs reading the env variables. A fixed version TestContex.new_from_env_new fixes that, but it does not work on with the build server.

HJLebbink:
@donatello 

I fixed that tests may not always cleanup their stuff. 

PTAL

donatello:
> TestContext.new_from_env (that reads the environment variables) only works with the configuration used by the build server. For a setup with play.min.io or my local box, it has bugs reading the env variables. A fixed version TestContex.new_from_env_new fixes that, but it does not work on with the build server.

For this part, you can check a CI variable like `CI=true` https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

donatello:
It is also ok to hardcode play creds for running tests in the local environment. I suggest something like:

1. if CI=true, assume the minio server launched in CI environment for running in github CI
2. Otherwise, look for endpoint env var like `MINIO_ENDPOINT=domain:port` with MINIO_ACCESS_KEY and MINIO_SECRET_KEY for creds (maybe default these to `minioadmin:minioadmin` if not present) and MINIO_SECURE=1 for https (default to http).
3. if no endpoint is given, assume play server's creds and run tests.

donatello (approved):

PR #133, opened 2025-03-12
Remove os_info dep

### LINKED ISSUE #107: os_info crate
Using the os_info create prevents to build docker images from scratch.
As far as I can see it is only for building a user_agent String. This can
be done in an other way, or make this via a feature configurable.

donatello:
os_info is too heavy, for the purposes of this library using the consts in std::env are enough

Fixes #107


HJLebbink (approved):
LGTM

PR #134, opened 2025-03-13
refactor make-bucket, bucket_exists, remove_bucket

HJLebbink:


src/s3/builders/make_bucket.rs
```diff
@@ -0,0 +1,151 @@
...
+        let client: &Client = self.client.as_ref().ok_or(Error::NoClientProvided)?;
+
+        let req = S3Request::new(client, Method::PUT)
+            .region(Some(region))
+            .bucket(Some(&self.bucket))
+            .query_params(query_params)
+            .headers(headers)
+            .body(body);
+
+        //TODO insert into region_map used to be execution after creating the bucket...
```
HJLebbink, line 144: @donatello Before the refactor, the `region_map` was updated after the payload (MinIO update) was executed. But in the current structure, there's nothing keeping track of when the `region_map` needs updating. Do you think it would cause any issues if we updated it before the payload? This `region_map` feels a bit awkward, are we sure we even need it?

PR #136, opened 2025-03-13
Wrap region headers map in an Arc

jhonboy121:
Client struct is not truly cheap to clone since DashMap clone is not cheap. Wrap it in an Arc to fix this.


HJLebbink:
@jhonboy121 I'm currently (as we speak) refactoring region_map. Give me a few days to sort things out.

HJLebbink (approved):
LGTM

PR #137, opened 2025-03-14
Remove deprecated downloads, uploads, and multipart API

donatello:
Add From<&'static str> for ObjectContent and remove deprecated API


HJLebbink (approved):
LGTM, that saves me a lot of refactoring!

PR #138, opened 2025-03-17
Refactor legal hold and make bucket

HJLebbink:
Refactored: bucket_exists, make_bucket, remove_bucket.
Refactored: enable_object_legal_hold, disable_object_legal_hold, is_object_legal_hold_enabled.

Additionally, some bugfixes, and added missing tests.


src/s3/builders/make_bucket.rs
```diff
@@ -0,0 +1,151 @@
...
+            if !base_url.region.is_empty() {
+                if base_url.region != *r {
+                    return Err(Error::RegionMismatch(
+                        base_url.region.clone(),
+                        r.to_string(),
+                    ));
+                }
+                region = r;
+            }
+        }
```
donatello, line 99: Perhaps we need some fix here: the region can be set via `MakeBucket::region` or by `base_url.region`. (If both are provided, they must be equal). If both are empty (or not given), we default to "us-east-1".

src/s3/client.rs
```diff
@@ -180,9 +186,9 @@ impl ClientBuilder {
 #[derive(Clone, Debug, Default)]
 pub struct Client {
     client: reqwest::Client,
-    base_url: BaseUrl,
+    pub base_url: BaseUrl,
```
donatello, line 189:
```suggestion
    pub(crate) base_url: BaseUrl,
```

src/s3/client.rs
```diff
@@ -180,9 +186,9 @@ impl ClientBuilder {
...
     provider: Option<Arc<Box<(dyn Provider + Send + Sync + 'static)>>>,
-    region_map: Arc<DashMap<String, String>>,
+    pub region_map: Arc<DashMap<String, String>>,
```
donatello, line 191:
```suggestion
    pub(crate) region_map: Arc<DashMap<String, String>>,
```

This prevents usage by non-minio-rs code.

donatello (changes requested):

src/s3/builders/make_bucket.rs
HJLebbink: I've updated the code. PTAL

PR #140, opened 2025-03-18
Refactor object lock config, retention, tags

HJLebbink:
Refactor object_lock_config,
Refactor object retention,
Refactor object tags,


HJLebbink:
@donatello This is a PR with all the refactoring combined, The remaining functions are slightly more complex and may take some more time. PTAL

src/s3/builders/make_bucket.rs
```diff
@@ -0,0 +1,155 @@
...
+        let client: &Client = self.client.as_ref().ok_or(Error::NoClientProvided)?;
+
+        let req = S3Request::new(client, Method::PUT)
+            .region(Some(region))
+            .bucket(Some(&self.bucket))
+            .query_params(query_params)
+            .headers(headers)
+            .body(body);
+
+        //TODO insert into region_map used to be executed after creating the bucket...
```
donatello, line 148: Can be removed?

src/s3/builders/set_object_retention.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub object: String,
+    pub version_id: Option<String>,
+    pub bypass_governance_mode: bool,
+    pub retention_mode: Option<RetentionMode>,
+    pub retain_until_date: Option<UtcTime>,
+}
+
+impl SetObjectRetention {
+    pub fn new(bucket: &str) -> Self {
+        //TODO make bucket of type String because its cloned anyway
```
donatello, line 44: Can remove.

donatello (commented):
Looks good, just wondering about TODOs some of which already seem to be done.

src/s3/builders/make_bucket.rs
HJLebbink: @donatello This change is the only functional one (possibly unintentional). Previously, the region map was updated after the MinIO update, but after the refactor, it's now updated before the MinIO update. If you think this doesn't matter or shouldn't, we can just remove the TODO.

src/s3/builders/set_object_retention.rs
HJLebbink: Everywhere, we currently have `fn new(bucket: &str) -> Self`, but I think `fn new(bucket: String) -> Self` would be better since it avoids an unnecessary clone when the caller no longer needs bucket. @donatello, what do you think?

donatello: No I think most of the caller would be reusing a bucket variable across multiple minio-rs library api calls. So it makes it sense to take a reference and clone internally. Additionally, it's easier to use a literal string as an argument like `client.make_bucket("mybucket")` this way.

src/s3/builders/make_bucket.rs
donatello: Ah i see, we dont know if the make bucket attempt succeeded, but we insert it into the region map anyway. Problem is by the we check the response, I think we do not  have the client object in hand to update the state. I think to be correct, we should not be setting it in the region cache at all. Let a subsequent request on the bucket insert it into the region cache.

src/s3/builders/make_bucket.rs
```diff
@@ -0,0 +1,155 @@
...
+        client
+            .region_map
+            .insert(self.bucket.clone(), region.to_string());
```
donatello, line 151: We can basically remove this.

HJLebbink: @donatello I have moved the region cache updates to the response handling. Turns out it was easier to implement than I first thought. I think the region caching is now as it is supposed to be. I have also sorted out a lot of the TODOs. 

Would you PTAL?

donatello (approved):

PR #141, opened 2025-03-19
optimize: use ring hash and hmac implementations

jhonboy121:
Ring has a faster hash implementation than sha2 crate. After profiling with samply and analysing flamegraph, hash computation call from sha256_hash_sb cpu usage went down to 2.6% from 13% just from this change.

Screenshot 2025-03-19 at 9 27 10 AM

Screenshot 2025-03-19 at 9 28 48 AM

### DIFF
--- a/src/s3/utils.rs
+++ b/src/s3/utils.rs
@@ -22,11 +22,15 @@ use base64::engine::general_purpose::STANDARD as BASE64;
 use byteorder::{BigEndian, ReadBytesExt};
 use chrono::{DateTime, Datelike, NaiveDateTime, ParseError, Utc};
 use crc::{CRC_32_ISO_HDLC, Crc};
+use hex::ToHex;
 use lazy_static::lazy_static;
 use md5::compute as md5compute;
 use multimap::MultiMap;
 use percent_encoding::{AsciiSet, NON_ALPHANUMERIC, percent_decode_str, utf8_percent_encode};
 use regex::Regex;
+#[cfg(feature = "ring")]
+use ring::digest::{Context, SHA256};
+#[cfg(not(feature = "ring"))]
 use sha2::{Digest, Sha256};
 pub use urlencoding::decode as urldecode;
 pub use urlencoding::encode as urlencode;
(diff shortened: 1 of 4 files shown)


HJLebbink (dismissed):
Thank you, LGTM

vnghia:
FYI, ring has been moved to minimal maintenance mode (https://github.com/briansmith/ring/discussions/2414). I think we should not use ring as a result

HJLebbink:
> FYI, ring has been moved to minimal maintenance mode ([briansmith/ring#2414](https://github.com/briansmith/ring/discussions/2414)). I think we should not use ring as a result

Absolutely, thanks for pointing that out.

jhonboy121:
@vnghia minimal maintenance mode does not mean its abandanoned. It is one of the well tested crates out there afaik, and there will be security patches as per the maintainer as and when that person or others can find time. Still, rust crates dont need to be constantly updated like other ecosystems, if it runs then it works just fine most of the cases. I have been performance testing my services and this patch makes a considerable difference in throughput. It would be great if it is merged with upstream, otherwise I will continue with my fork. Thanks!

vnghia:
> It is one of the well tested crates out there afaik, and there will be security patches as per the maintainer as and when that person or others can find time. Still, rust crates dont need to be constantly updated like other ecosystems, if it runs then it works just fine most of the cases.

Yeah I totally agree with you in this point. And I myself will happily use `ring` to squeezing any bit of performance in my personal project. However, I think some people will be a little bit more sensitive because `ring` has a lot of things to do with security.

How about gating it behind a cargo feature ? What do you think about it @jhonboy121 ? Thank you for improving the library!

jhonboy121:
@vnghia making ring opt-in is a great idea, will submit a patch for that.

src/s3/utils.rs
```diff
@@ -68,17 +69,15 @@ pub fn uint32(mut data: &[u8]) -> Result<u32, std::io::Error> {
...
 }
 
 pub fn sha256_hash_sb(sb: &SegmentedBytes) -> String {
-    let mut hasher = Sha256::new();
+    let mut context = Context::new(&SHA256);
     for data in sb.iter() {
-        hasher.update(data);
+        context.update(data.as_ref());
     }
-    format!("{:x}", hasher.finalize())
```
vnghia, line 81: Also AFAIK this is not a great way to convert to hex string. I personally avoid `format!` as much as possible and in this particular case, I would either use:
- `hex::encode` since we already have it as a dependency.
- `const_hex::encode` new shiny kid that is 10x~50x faster than `hex`.

jhonboy121:
@vnghia @HJLebbink updated PR, ring will be opt in. Defaults to sha2 and hmac.

src/s3/utils.rs
jhonboy121: Correct, this was in upstream and I used hex crate to encode in my patch since it was already a dependency. TIL about const_hex, will check it out. Thanks!

vnghia (approved):
LGTM. Thank you!

HJLebbink (approved):
LGTM, thank you!

PR #144, opened 2025-03-25
added `append_object` from express-zone

HJLebbink:
For the tests to pass, the express-zone needs to be merged into EOS first. See https://github.com/miniohq/eos/pull/572, and related minio-go SDK PR https://github.com/minio/minio-go/pull/2082


harshavardhana:
The API is not called AppendPart() instead since this API has nothing to do with multipart API.

We should simply call this AppendObject() and perform the Stat() call ourselves internally to provide the correct object offset and then simply take an input Reader stream.

We shouldn't expect users to provide the right offset, what we can however do is take how much data will appended() per AppendObject() call as a stream.

Additionally we should provide client a way to chunk the stream, default add the entire stream as single call or perform configured chunks worth append() can be any user configured value.

HJLebbink:
@harshavardhana @donatello I have added support for streaming appends. The five append tests are currently disabled, but they run successfully on my local box. There's one unrelated test failing for bucket encryption. not sure what's causing that. PTAL

harshavardhana:
> @harshavardhana @donatello I have added support for streaming appends. The five append tests are currently disabled, but they run successfully on my local box. There's one unrelated test failing for bucket encryption, if the --express-zone is switched off the issues disappear. not sure what's causing that. PTAL

Thats because SSE is not supported yet for append() API - in the test case however you should look for the error response as "NotSupported" then skip it.

HJLebbink:
> > @harshavardhana @donatello I have added support for streaming appends. The five append tests are currently disabled, but they run successfully on my local box. There's one unrelated test failing for bucket encryption, if the --express-zone is switched off the issues disappear. not sure what's causing that. PTAL
> 
> Thats because SSE is not supported yet for append() API - in the test case however you should look for the error response as "NotSupported" then skip it.

@harshavardhana Thanx for the definitive answers.

HJLebbink:
This PR is subsumed under https://github.com/minio/minio-rs/pull/143

PR #145, opened 2025-04-01
minio-rs: refactored all functions

HJLebbink:
refactored stat_object
refactored select_object_content
refactor get_presigned_object_url
refactor get_presigned_policy_form_data
refactored upload-part-copy


HJLebbink:
@donatello I noticed a few functional differences between the various builders. I've tried to keep changes to a minimum, but this one really stands out and feels like a bug.

Some functions (like get_bucket_encryption) use the region provided by the user directly when sending an S3Request.

```rust
impl ToS3Request for GetBucketEncryption {
    fn to_s3request(self) -> Result<S3Request, Error> {
        check_bucket_name(&self.bucket, true)?;

        Ok(S3Request::new(self.client, Method::GET)
            .region(self.region)   // HERE
            .bucket(Some(self.bucket))
            .query_params(insert(self.extra_query_params, "encryption"))
            .headers(self.extra_headers.unwrap_or_default()))
    }
}
```

Whereas other functions (like get_bucket_lifecycle) use the region provided by the user to look up a cached region (and if it is not cached, makes a s3API get_region call)

```rust
impl ToS3Request for GetBucketLifecycle {
    fn to_s3request(self) -> Result<S3Request, Error> {
        check_bucket_name(&self.bucket, true)?;

        let region: String = self.client.get_region_cached(&self.bucket, &self.region)?;

        Ok(S3Request::new(self.client, Method::GET)
            .region(Some(region))
            .bucket(Some(self.bucket))
            .query_params(insert(self.extra_query_params, "lifecycle"))
            .headers(self.extra_headers.unwrap_or_default()))
    }
}
```

I reckon that for calls to minio it probably doesn't matter, but since we should be S3 compatible, I assume we need to do the cached lookup in all the builders? Right, or can we just drop the cached lookup accross the board?

donatello (approved):

PR #157, opened 2025-05-22
Add crates.io link to README

vnghia:


HJLebbink:
@vnghia Cheers for the effort, much appreciated! We are in the middle of rewriting the README and docs at the moment, so hang tight and keep an eye out for updates. Would love your input once things have settled a bit!

PR #160, opened 2025-05-26
Fixed doc warnings; fixed minor issue with trait separation in append_object

HJLebbink:
Fixed a minor issue with object name and bucket name impl trait separation

### DIFF
--- a/src/s3/builders/delete_objects.rs
+++ b/src/s3/builders/delete_objects.rs
@@ -31,7 +31,7 @@ use crate::s3::utils::{check_object_name, insert};
 use crate::s3::{
     Client,
     error::Error,
-    response::{RemoveObjectResponse, RemoveObjectsResponse},
+    response::{DeleteObjectResponse, DeleteObjectsResponse},
     types::{S3Api, S3Request, ToS3Request, ToStream},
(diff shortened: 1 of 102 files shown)


src/s3/builders/delete_objects.rs
```diff
@@ -303,15 +309,19 @@ where
     }
 }
 
-// endregion: delete-objects
+// endregion: object-stream
 
-// region: remove-objects
+// region: remove-objects-steaming
```
donatello, line 314: Spelling typo. what is this "region" markup needed for?
```suggestion
// region: remove-objects-streaming
```

donatello (changes requested):
Please bump up the version in Cargo.toml to 0.3.0

src/s3/builders/delete_objects.rs
HJLebbink: Fixed,  The `// region` with `// endregion` is to create a folding region, and to keep me sane and organized...

image

HJLebbink:
@donatello Yes, bumped the version, Should be ready now!

donatello (approved):

PR #162, opened 2025-06-01
Duplicated code removed, and lazy response evaluation

HJLebbink:
This PR tackles a lot of duplicated code in the response types. For example, the same bit of code to extract the bucket from a request was repeated 45 times, that’s now been replaced with a HasBucket trait. Many other such Has traits have been defined.

Another key improvement is that responses are now lazily evaluated. Previously, the response would parse the entire response body up front, even if none of the properties were actually needed by the user. With this PR, response properties are only fetched when they’re actually requested.


donatello (approved):
LGTM, need to bump up 0.4.0 before next release.

PR #164, opened 2025-06-16
fix: replace Arc<Box<...>> credential-provider with Arc<...>

twuebi:
A `Box` is a pointer and an `Arc` is a smart pointer, we can remove one layer of indirection by directly using `Arc`. 
We can also remove the need for our caller to box / arc the input by making the respective functions generic over `Provider`.


HJLebbink (approved):
LGTM, works as advertised

PR #167, opened 2025-06-17
moved Tokio runtime from general dependency to dev dependency

HJLebbink:


src/s3/error.rs
```diff
@@ -129,7 +129,10 @@ pub enum Error {
     StrError(reqwest::header::ToStrError),
     IntError(std::num::ParseIntError),
     BoolError(std::str::ParseBoolError),
-    Utf8Error(alloc::string::FromUtf8Error),
+    /// Occurs when converting Vec<u8> to String (e.g. String::from_utf8)
+    FromUtf8Error(alloc::string::FromUtf8Error),
+    /// Occurs when converting &[u8] to &str (e.g. std::str::from_utf8)
+    Utf8Error(std::str::Utf8Error),
```
twuebi: Is this distinction meaningful to our users? We could have one `Utf8Error(Box<dyn std::error::Error + Send + Sync + 'static>)` instead.

src/s3/object_content.rs
```diff
@@ -203,7 +216,7 @@ impl ObjectContent {
             ))
         })?;
         if !parent_dir.is_dir() {
-            fs::create_dir_all(parent_dir).await?;
+            fs::create_dir_all(parent_dir)?;
```
twuebi: I think this is a blocking fs call which can lead to blocking the executor (https://ryhl.io/blog/async-what-is-blocking/), could we use https://docs.rs/async-std/latest/async_std/fs/fn.create_dir_all.html instead?

src/s3/object_content.rs
```diff
@@ -219,19 +232,18 @@ impl ObjectContent {
...
                 break;
             }
             total_bytes_written += bytes.len() as u64;
-            fp.write_all(&bytes).await?;
+            fp.write_all(&bytes)?;
         }
-        fp.flush().await?;
-        fs::rename(&tmp_file_path, file_path).await?;
+        fp.flush()?;
+        fs::rename(&tmp_file_path, file_path)?;
```
twuebi: Same as above, I think we should use async-std's api here too if we want to get rid of the tokio dependency https://docs.rs/async-std/latest/async_std/fs/struct.File.html

src/s3/client.rs
```diff
@@ -268,47 +266,40 @@ impl Client {
...
                         }
+                    } else {
                         false
-                    });
-                    self.shared.express.set(express).unwrap_or_default();
-                    express
+                    }
+                }
+                Err(e) => {
+                    println!("is_express_internal: error: {e}\nassume false");
```
twuebi: let's use `log::warn!` or `log::info!` here

tests/test_bucket_replication.rs
```diff
@@ -29,7 +29,7 @@ use minio_common::test_context::TestContext;
 #[tokio::test(flavor = "multi_thread", worker_threads = 10)]
```
twuebi: do you know why all these tests use 10 worker threads?

src/s3/error.rs
HJLebbink: Yes, makes sense.

twuebi (approved):
I feel we should revisit the test cleanup at some point in the future

twuebi (approved):

PR #168, opened 2025-06-17
wip: minio client-hooks

twuebi:
Adds a `ClientHooks` trait which can be used to hook into certain points in a request life-cycle. For now, it only adds two hooks but we may add more hooks in the future as we need them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ version = "0.12.18"
 default-features = false
 features = ["stream"]
 
+
 [features]
 default = ["default-tls", "dep:hmac", "dep:sha2"]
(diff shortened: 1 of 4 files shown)


HJLebbink:
@twuebi Can we merge this functionality into the main branch? Are we missing anything?

twuebi:
> @twuebi Can we merge this functionality into the main branch? Are we missing anything?

Since it's a public facing API we may wanna think a bit of other hooks / names for things since they will be hard to change once released, otherwise, I think we should be good, I'll do another pass tomorrow morning.

HJLebbink:
superseded by [PR #188](https://github.com/minio/minio-rs/pull/188)

PR #169, opened 2025-06-18
Removed `futures` dependency

alkarkhi:
The dependency is not needed since `futures_util` does everything it's used for


HJLebbink (approved):
LGTM

HJLebbink:
Thanks for reporting this! I’ve added your changes and cleaned up other unnecessary includes in a newer PR, so this one is now covered.

PR #170, opened 2025-06-18
feat: add a test macro

twuebi:
1. Added a test macro `#[minio_macros::test]`, it performs setup and tear down for tests (currently creation + deletion of bucket), if no bucket is wished `#[minio_macros::test(no_bucket)]` will be your friend. If the test panics, it'll catch the unwind and attempt cleanup before resuming the unwind.
2. replace some blocking io in tests with async_std variants

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -36,7 +36,7 @@ jobs:
           export ACCESS_KEY=minioadmin
           export SECRET_KEY=minioadmin
           export ENABLE_HTTPS=1
-          export SSL_CERT_FILE=./tests/public.crt
+          export MINIO_SSL_CERT_FILE=./tests/public.crt
           MINIO_TEST_TOKIO_RUNTIME_FLAVOR="multi_thread" cargo test -- --nocapture
   test-current-thread:
     runs-on: ubuntu-latest
@@ -49,7 +49,7 @@ jobs:
           export ACCESS_KEY=minioadmin
           export SECRET_KEY=minioadmin
           export ENABLE_HTTPS=1
-          export SSL_CERT_FILE=./tests/public.crt
+          export MINIO_SSL_CERT_FILE=./tests/public.crt
           MINIO_TEST_TOKIO_RUNTIME_FLAVOR="current_thread" cargo test -- --nocapture
 
   build:
--- a/benches/s3/common_benches.rs
+++ b/benches/s3/common_benches.rs
@@ -40,7 +40,7 @@ impl Ctx2 {
     /// Create a new context with a bucket
     pub async fn new() -> Self {
         unsafe {
-            env::set_var("SSL_CERT_FILE", "./tests/public.crt");
+            env::set_var("MINIO_SSL_CERT_FILE", "./tests/public.crt");
         }
         let ctx = TestContext::new_from_env();
         let (bucket_name, cleanup) = ctx.create_bucket_helper().await;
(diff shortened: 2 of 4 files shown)


twuebi:
I'm still tracking down some functions that seem to take too long between await points leading to some really slow/hanging tests when running on the standard tokio runtime (https://ryhl.io/blog/async-what-is-blocking/). I think our `impl Stream for RandSrc` may be the culprit, calls to poll_next take anything between 200 - 800 microseconds, will have to look a bit more into it.

macros/src/lib.rs
```diff
@@ -0,0 +1,92 @@
+mod test_attr;
```
HJLebbink, line 15: Could you please add the Apache License to this and other code files. I think some find it necessary.

macros/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+/// #[minio_macros::test(no_bucket)]
+/// async fn my_test(ctx: TestContext) {
+///    // Your test code here
+/// }
+///```
+/// The macro also supports additional arguments:
+///
+/// - `flavor`: Specifies the flavor of the Tokio test (e.g., "multi_thread").
+/// - `worker_threads`: Specifies the number of worker threads for the Tokio test.
+/// - `bucket_name`: Specifies the name of the bucket to be used in the test. If not provided, a random bucket name will be generated.
```
HJLebbink, line 57: Missing line doc for `object_lock` somewhere around here

macros/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "minio-macros"
+version = "0.1.0"
+edition = "2021"
```
HJLebbink: Im not sure whether it matters, but would using 2024 here restrict the project to be used in 2024 edition only?

common/src/rand_src.rs
```diff
@@ -41,26 +40,21 @@ impl Stream for RandSrc {
...
-            return task::Poll::Ready(None);
+            return Poll::Ready(None);
         }
-
-        let bytes_read = match self.size > 64 * 1024 {
-            true => 64 * 1024,
-            false => self.size as usize,
-        };
+        // Limit to 8 KiB per read
+        let bytes_read = self.size.min(8 * 1024) as usize;
```
twuebi, line 49: at 64kb this method takes ~300-500micros which is a bit long

src/s3/client.rs
```diff
@@ -424,7 +424,7 @@ impl Client {
         query_params: &Multimap,
         bucket_name: Option<&str>,
         object_name: Option<&str>,
-        body: Option<&SegmentedBytes>,
+        body: Option<Arc<SegmentedBytes>>,
```
HJLebbink, line 427: In the past, we have bundled all sorts of changes: doc updates, test tweaks, bug fixes, and core functionality, into a single PRs. This change is a bit more involved than just modifying tests.

I would like to fully understand the impact on other methods that only have a single segment, which covers about 99% of cases. I don't like to rush this change. That said, all the performance-critical functions will deal with multiple segments anyway.

Just to check: is this change actually necessary for the tests to run properly?

macros/Cargo.toml
donatello: I think we should use 2024. It increases minimum version of the compiler needed to compile this package.

donatello (approved):
LGTM - thanks for adding the current_thread/multi_thread tests!

src/s3/client.rs
twuebi: It's put behind an Arc since we are doing a spawn_blocking capturing it when hashing, without the Arc, we'd be copying. We do spawn_blocking since the payload can be fairly large.

PR #177, opened 2025-07-10
Tests cleanup; cargo clippy fixes, minor doc updates

HJLebbink:
Minor issues found while debugging an issue

### DIFF
--- a/src/s3/error.rs
+++ b/src/s3/error.rs
@@ -44,7 +44,7 @@ pub enum ErrorCode {
     ResourceConflict,
     AccessDenied,
     NotSupported,
-    BucketNotEmpty,
+    BucketNotEmpty(String), // String contains optional reason msg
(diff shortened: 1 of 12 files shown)


src/s3/error.rs
```diff
@@ -75,7 +75,7 @@ impl ErrorCode {
             "resourceconflict" => ErrorCode::ResourceConflict,
             "accessdenied" => ErrorCode::AccessDenied,
             "notsupported" => ErrorCode::NotSupported,
-            "bucketnotempty" => ErrorCode::BucketNotEmpty,
+            "bucketnotempty" => ErrorCode::BucketNotEmpty("".to_string()),
```
twuebi, line 78: could make that a Vec<String> instead of a string?

twuebi: or an option? Not a fan of empty string as default

Issue #165, opened 2025-06-16
Empty host can lead to runtime panics


twuebi:
the display implementation of the url type leads to runtime panics when the host is empty:

<img width="880" alt="Image" src="https://github.com/user-attachments/assets/c53255d0-fcce-4905-850e-4128c92a4c8d" />

```
thread 'main' panicked at library/std/src/io/mod.rs:648:17:
a formatting trait implementation returned an error when the underlying stream did not
stack backtrace:
```
 
`ClientBuilder::default()` constructs a client with empty host.

PR #182, opened 2025-08-15
fixed clippy issues and cleanup of unnecessary imports

HJLebbink:


src/s3/http.rs
```diff
@@ -58,6 +55,18 @@ impl Url {
     }
 }
 
+impl Default for Url {
+    fn default() -> Self {
+        Self {
+            https: true,
+            host: String::default(),
```
twuebi, line 62: I think this default may crash on display: https://github.com/minio/minio-rs/issues/165

src/s3/utils.rs
```diff
@@ -88,32 +87,51 @@ pub const EMPTY_SHA256: &str = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934c
...
+        v.set_len(len * 2);
+        for i in 0..len {
+            let b = bytes.get_unchecked(i);
+            let hi = LUT.get_unchecked((b >> 4) as usize);
+            let lo = LUT.get_unchecked((b & 0xF) as usize);
+            let j = i * 2;
+            *v.get_unchecked_mut(j) = *hi;
+            *v.get_unchecked_mut(j + 1) = *lo;
+        }
+    }
```
twuebi: Maybe put a safety comment which explains why this unsafe block is sound (and needed)

twuebi (changes requested):
Looks mostly good, let's check the display implementation on BaseUrl & Url when constructing the default & add a small [safety comment](https://std-dev-guide.rust-lang.org/policy/safety-comments.html) on the unsafe block

src/s3/http.rs
HJLebbink: I changed it to
```
host: "127.0.0.1".to_string(),
port: 9000,
```
I guess that would solve that crash,

twuebi: I think we should rather fix the display to not crash at runtime on an empty host. Why do we want a default for Url at all?

HJLebbink: The old code had a default value true, and to keep that I added a hand-made default impl, this is the delta:
image

because `..Default::default()` is used at many places in the builders.

But yeah, better to fix the crash, when I have time.

HJLebbink: Will fix this in the next upcoming update with the cleanup of the defaults

PR #184, opened 2025-09-10
use crate TypedBuilder for builders

HJLebbink:


tests/test_object_retention.rs
```diff
@@ -52,8 +53,9 @@ async fn object_retention(ctx: TestContext, bucket_name: String) {
     let resp: PutObjectRetentionResponse = ctx
         .client
         .put_object_retention(&bucket_name, &object_name)
-        .retention_mode(Some(RetentionMode::GOVERNANCE))
-        .retain_until_date(Some(retain_until_date))
+        .retention_mode(RetentionMode::GOVERNANCE)
+        .retain_until_date(retain_until_date)
```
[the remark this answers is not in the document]
HJLebbink: Untrue, you can still call with Some(RetentionMode::GOVERNANCE)

tests/test_object_put.rs
```diff
@@ -156,7 +164,8 @@ async fn put_object_content_2(ctx: TestContext, bucket_name: String) {
                 &object_name,
                 ObjectContent::new_from_stream(data_src, None),
             )
-            .part_size(Some(MIN_PART_SIZE))
+            .part_size(Known(MIN_PART_SIZE))
```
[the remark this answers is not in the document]
HJLebbink: fixed

src/s3/client/delete_bucket.rs
```diff
@@ -111,31 +117,34 @@ impl Client {
...
             }
         }
 
-        let request: DeleteBucket = self.delete_bucket(&bucket);
+        let request: DeleteBucket = self.delete_bucket(&bucket).build();
         match request.send().await {
             Ok(resp) => Ok(resp),
             Err(Error::S3Server(S3Error(mut e))) => {
                 if matches!(e.code(), MinioErrorCode::NoSuchBucket) {
+                    let dummy: S3Request = S3Request::builder().client(self.clone()).method(Method::DELETE).bucket(bucket).headers(MultiMap::default()).build(/* S3RequestBuilder_Error_Missing_required_field_headers */);
```
twuebi: What's with the comment here?

src/s3/client.rs
```diff
@@ -120,22 +120,32 @@ pub const MAX_OBJECT_SIZE: u64 = 5_497_558_138_880; // 5 TiB
...
+pub struct MinioClientBuilder {
+    //#[builder(!default)] // force required
     base_url: BaseUrl,
+    //#[builder(default, setter(into, doc = "Set the credential provider. If not, set anonymous access is used."))]
     provider: Option<Arc<dyn Provider + Send + Sync + 'static>>,
+    //#[builder(default, setter(into, doc = "Set file for loading CAs certs to trust. This is in addition to the system trust store. The file must contain PEM encoded certificates."))]
     ssl_cert_file: Option<PathBuf>,
+    //#[builder(default, setter(into, doc = "Set flag to ignore certificate check. This is insecure and should only be used for testing."))]
     ignore_cert_check: Option<bool>,
+    //#[builder(default, setter(into, doc = "Set the app info as an Option of (app_name, app_version) pair. This will show up in the client's user-agent."))]
```
twuebi: remove those commented lines?

src/s3/client.rs
```diff
@@ -491,13 +506,17 @@ impl Client {
...
-                body.as_ref().unwrap()
+                header_strings.join("; ")
             );
+            let truncated = if debug_str.len() > 1000 {
+                format!("{}...", &debug_str[..997])
+            } else {
+                debug_str
+            };
+            println!("{truncated}");
         }
```
twuebi, line 521: let's remove this instead? It's hidden behind a if false, no need to have this checked in. We should add [tracing](https://docs.rs/tracing/latest/tracing/) to our dependencies and then debug logs can be easily turned on/off via RUST_LOG env.

src/s3/client.rs
```diff
@@ -591,9 +610,23 @@ impl Client {
         )
         .await
     }
+
+    /// create an example client for testing on localhost
+    pub fn create_client_on_localhost()
```
twuebi, line 617: let's feature-gate this, it's not useful for anyone using this beyond examples

src/s3/http.rs
twuebi: This is the cause of #165, we should avoid erroring inside fmt calls, it'll lead to runtime panics

README.md
```diff
@@ -24,8 +24,8 @@ use minio::s3::response::BucketExistsResponse;
 
 #[tokio::main]
 async fn main() {
-    let client: Client = Default::default(); // configure your client
-
+    let client = Client::create_client_on_localhost().unwrap(); // configure your client here
```
twuebi: `s/Client/MinioClient/`, right?

twuebi (changes requested):
some nits, mostly good!

src/s3/client/delete_bucket.rs
HJLebbink: good question, its not visible, would need to scroll to see it, so i don't know... Let me fix that.

src/s3/client.rs
HJLebbink: This is there for when the hook stuff gets merged, someday. I had expected it to stay there for 24 hours...

src/s3/client.rs
HJLebbink: Good idea. Let me fix that.

src/s3/client.rs
HJLebbink: I vaguely remember trying to migrate the handcrafted ClientBuilder struct to the builder infrastructure that was added in this commit. I would need to dig into it again to fully understand why that was not possible. Those commented lines were doing just that: creating a generated MinIO client builder.

src/s3/client.rs
twuebi: I saw we already got log as a dependency, maybe just replace it with a. log::debug!? Also does headers contain sensitive information?

HJLebbink: I will have this fixed once https://github.com/minio/minio-rs/pull/168 is done. I'd rather put time into finishing that PR instead of making this one bigger...

PR #188, opened 2025-10-17
feat: add request hooks infrastructure

HJLebbink:
Adds RequestHooks trait to enable intercepting and modifying S3 API
  requests at key points in the request lifecycle. This enables implementing 
  cross-cutting concerns like load balancing, telemetry, and debug logging
  without modifying the core request handling logic.

  Key features:
  - before_signing_mut: Modify URL, headers before request signing
  - after_execute: Inspect responses for telemetry/logging
  - x-minio-redirect header: Automatically added when URL is modified to
    inform servers about client-side redirections
  - Extensions: Type-safe state bag for passing data between hooks

  Implementation details:
  - Hooks run in order they were added to client builder
  - Errors in before_signing_mut abort the request
  - Errors in after_execute are logged but don't fail the request
  - Removed old disabled debug logging code, replaced with hook pattern

  Added comprehensive documentation:
  - Module-level docs explaining x-minio-redirect header purpose
  - Trait documentation with load balancing and telemetry examples
  - Inline comments explaining client-side redirection behavior

  Testing:
  - 6 unit tests covering URL modification, header modification, extensions,
    error handling, and default implementations
  - New example: debug_logging_hook.rs demonstrating hook usage
  - All existing tests pass, clippy warning-free

  Example usage:
  ```rust
  let debug_hook = Arc::new(DebugLoggingHook::new(true));
  let client = MinioClientBuilder::new(base_url)
      .hook(debug_hook)
      .build()?;
```
  Related optimizations:
  - Removed unnecessary body.clone() in request handling
  - Removed unused ReqwestError import
  - Optimized URL comparison to avoid string allocations

  This commit message:
  - Follows conventional commit format (feat:)
  - Describes what was added and why
  - Lists key features
  - Explains implementation details
  - Documents what was removed/replaced
  - Mentions testing and examples
  - Includes a usage snippet
  - Notes related optimizations

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -34,14 +34,14 @@ All source files that haven't been generated MUST include the following copyrigh
 
 ### Comments
 - **NO redundant comments** - Code should be self-documenting
-- Avoid obvious comments like `// Set x to 5` for `x := 5`
+- Avoid obvious comments like `// Set x to 5` for `let x = 5;`
 - Only add comments when they explain WHY, not WHAT
 - Document complex algorithms or non-obvious business logic
 
 ## Critical Code Patterns
 
 ### Builder Pattern
-All S3 API requests MUST use the builder pattern, with the following documentation but then for the appropriate API
+All S3 API requests MUST use the builder pattern, with documentation similar to the following example (adapted for each specific API)
 
 ```rust
 /// Argument builder for the [`AppendObject`](https://docs.aws.amazon.com/AmazonS3/latest/userguide/directory-buckets-objects-append.html) S3 API operation.
@@ -72,6 +72,33 @@ impl Client {
     }
 }
 ```
+
+### Rust-Specific Best Practices
+
+1. **Ownership and Borrowing**
+   - Prefer `&str` over `&String` in function parameters
+   - Use `AsRef<str>` or `Into<String>` for flexible string parameters
+   - Return owned types from functions unless lifetime annotations are clear
+
+2. **Type Safety**
+   - Use `#[must_use]` attribute for functions returning important values
+   - Prefer strong typing over primitive obsession
+   - Use newtypes for domain-specific values
+
+3. **Unsafe Code**
+   - Avoid `unsafe` code unless absolutely necessary
+   - Document all safety invariants when `unsafe` is required
+   - Isolate `unsafe` blocks and keep them minimal
+
+4. **Performance**
+   - Use `Cow<'_, str>` to avoid unnecessary allocations
+   - Prefer iterators over collecting into intermediate vectors
+   - Use `Box<dyn Trait>` sparingly; prefer generics when possible
+
+5. **Async Patterns**
+   - Use `tokio::select!` for concurrent operations
+   - Avoid blocking operations in async contexts
+   - Use `async-trait` for async trait methods
 
 ## Code Quality Principles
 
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: examples/debug_logging_hook.rs [hunk 1]
```diff
@@ -0,0 +1,235 @@
...
+        _body: Option<&SegmentedBytes>,
+        _extensions: &mut Extensions,
+    ) -> Result<(), Error> {
+        if self.verbose {
+            let bucket_obj = match (bucket_name, object_name) {
+                (Some(b), Some(o)) => format!("{b}/{o}"),
+                (Some(b), None) => b.to_string(),
+                _ => url.to_string(),
+            };
+            println!("→ Preparing {method} request for {bucket_obj}");
```
twuebi, line 91: let's use log or tracing instead of println

examples/debug_logging_hook.rs
```diff
@@ -0,0 +1,235 @@
...
+        // Format response status
+        let status = match resp {
+            Ok(response) => format!("✓ {}", response.status()),
+            Err(err) => format!("✗ Error: {err}"),
+        };
+
+        println!("━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━");
+        println!("S3 Request: {method} {bucket_obj}");
+        println!("URL: {url}");
+        println!("Status: {status}");
```
twuebi, line 124: let's use log or tracing instead

examples/debug_logging_hook.rs
twuebi: nvm, we're in an example

examples/debug_logging_hook.rs
twuebi: nvm, we're in an example

src/s3/client/hooks.rs
```diff
@@ -0,0 +1,517 @@
...
+/// ```ignore
+/// impl RequestHooks for TelemetryHook {
+///     async fn after_execute(&self, ..., resp: &Result<Response, ...>, ...) {
+///         // Log request metrics, duration, errors, etc.
+///     }
+/// }
+/// ```
+///
+/// ## Debug Logging
+/// ```ignore
```
twuebi: why ignore? These examples could be validated by CI otherwise.

src/s3/client.rs
```diff
@@ -441,55 +455,79 @@ impl MinioClient {
...
+        )
+        .await?;
+
+        // If a hook modified the URL (e.g., redirecting to a different MinIO node for load balancing),
+        // add the x-minio-redirect header to inform the server about the client-side redirection.
+        // This enables server-side telemetry, debugging, and load balancing metrics.
+        // The header contains the actual endpoint where the request is being sent.
+        if url.to_string() != url_before_hook {
+            headers.add("x-minio-redirect", url.to_string());
         }
```
twuebi, line 513: I'm wondering if sending the original URL would also be nice

src/s3/client/hooks.rs
HJLebbink: Yeah, should have been 'no_run'

src/s3/client.rs
HJLebbink: Yes, makes sense:  when a hook modifies the URL, the client now adds `x-minio-redirect-from` and `x-minio-redirect-to` headers for server-side tracking.

CLAUDE.md
twuebi, line 1: Do you intend to check this file in?

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.88.0"
+channel = "1.90.0"
 components = ["clippy", "rustfmt"]
 #targets = ["x86_64-unknown-linux-musl"]
```
twuebi: while we're here, let's remove the targets line

CLAUDE.md
HJLebbink: I copied the approach from other internal projects, so I suppose the answer is yes, but only because it saves others the time of creating one themselves.

rust-toolchain.toml
HJLebbink: fixed

PR #192, opened 2025-10-30
 Removed unnecessary intermediate Vec<Bytes> allocation in HTTP request body

HJLebbink:
## Summary

  - Removed unnecessary intermediate `Vec<Bytes>` allocation in HTTP request body preparation
  - Directly consume `SegmentedBytes` iterator into stream using `Arc::unwrap_or_clone()`
  - Implemented zero-cost `BodyIterator` enum to avoid heap allocation and dynamic dispatch
  - Reduces memory overhead and eliminates Vec growth reallocations during request preparation

## Memory Impact Analysis

  Important: Bytes is reference-counted, so cloning it never copies the actual byte data - just increments a reference count. The 1GB of actual data is
  never duplicated in either implementation. The difference is in metadata overhead - the collection of Bytes pointers.

  Scenario 1: Multipart Upload (5MB parts)

  For a 1GB file with MIN_PART_SIZE = 5 MiB:
  - Number of parts: 1,024 MB ÷ 5 MB = ~205 parts
  - Each Bytes pointer: 32 bytes (on 64-bit systems)

  Original Implementation:
  - Creates Vec<Bytes> with 205 entries
  - Memory overhead: 205 × 32 = 6,560 bytes ≈ 6.4 KB
  - Plus Vec's capacity over-allocation (typically 1.5-2x during growth)
  - Total: ~10-13 KB

  New Implementation:
  - BodyIterator enum: 16 bytes
  - Iterator state (indices): ~24 bytes
  - Total: ~40 bytes

  Savings: ~10-13 KB

  Scenario 2: Streaming Read (smaller chunks)

  If the 1GB is read from a stream in smaller buffers (e.g., 64KB chunks common in I/O):
  - Number of chunks: 1,048,576 KB ÷ 64 KB = 16,384 chunks

  Original Implementation:
  - Vec<Bytes> with 16,384 entries
  - Memory overhead: 16,384 × 32 = 524,288 bytes = 512 KB
  - With Vec growth: ~750 KB - 1 MB

  New Implementation:
  - Still ~40 bytes

  Savings: ~750 KB - 1 MB

  Additional Benefits

  1. Zero-cost abstraction - Enum dispatch is optimized away at compile time through monomorphization
  2. No heap allocation - Unlike Box<dyn Iterator>, the enum is stack-allocated
  3. No reallocation overhead - Vec growth causes multiple allocations and copies during .collect()
  4. Better cache locality - Preserves the nested SegmentedBytes structure which may already be in cache
  5. Lower peak memory - No temporary spike when collecting into the intermediate Vec

  Technical Details

  Arc::unwrap_or_clone() efficiently handles the Arc:
  - If sole owner: unwraps with zero cost
  - If shared: clones only the SegmentedBytes structure (Vec of Vecs), not the underlying byte data

  The BodyIterator enum provides type unification without runtime overhead, as the enum variants are resolved at compile time.

### DIFF
--- a/src/s3/client/mod.rs
+++ b/src/s3/client/mod.rs
@@ -119,6 +119,24 @@ pub const MAX_PART_SIZE: u64 = 5_368_709_120; // 5 GiB
 /// ensuring the object does not exceed 5 TiB.
 pub const MAX_OBJECT_SIZE: u64 = 5_497_558_138_880; // 5 TiB
 
+enum BodyIterator {
+    Segmented(crate::s3::segmented_bytes::SegmentedBytesIntoIterator),
+    FromVec(std::vec::IntoIter<Bytes>),
+    Empty(std::iter::Empty<Bytes>),
+}
+
+impl Iterator for BodyIterator {
+    type Item = Bytes;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self {
+            Self::Segmented(iter) => iter.next(),
+            Self::FromVec(iter) => iter.next(),
+            Self::Empty(iter) => iter.next(),
+        }
+    }
+}
+
 /// Maximum number of parts allowed in a multipart upload.
 ///
 /// Multipart uploads are limited to a total of 10,000 parts. If the object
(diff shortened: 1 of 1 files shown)


src/s3/client.rs
```diff
@@ -521,14 +521,11 @@ impl MinioClient {
...
 
         if (*method == Method::PUT) || (*method == Method::POST) {
-            //TODO: why-oh-why first collect into a vector and then iterate to a stream?
-            let bytes_vec: Vec<Bytes> = match body {
-                Some(v) => v.iter().collect(),
-                None => Vec::new(),
+            let iter: Box<dyn Iterator<Item = Bytes> + Send> = match body {
+                Some(v) => Box::new(Arc::unwrap_or_clone(v).into_iter()),
+                None => Box::new(std::iter::empty()),
             };
```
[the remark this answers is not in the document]
HJLebbink: Thanx

src/s3/client/mod.rs
```diff
@@ -521,14 +537,11 @@ impl MinioClient {
...
-                Some(v) => v.iter().collect(),
-                None => Vec::new(),
+            let iter = match body {
+                Some(v) => BodyIterator::Segmented(Arc::unwrap_or_clone(v).into_iter()),
+                None => BodyIterator::Empty(std::iter::empty()),
             };
-            let stream = futures_util::stream::iter(
-                bytes_vec.into_iter().map(|b| -> Result<_, Error> { Ok(b) }),
-            );
+            let stream = futures_util::stream::iter(iter.map(|b| -> Result<_, Error> { Ok(b) }));
```
twuebi, line 575: I may be missing something, we're still loading the whole body into memory within SegmentedBytes, wouldn't it make sense to have that already as a Stream so that e.g. uploading a file would not have to load the whole thing?

HJLebbink: The existing implementation is already fairly efficient, since nothing actually gets copied. The bytes are split into a vector of byte segments, it might look like data is being copied, but that is not the case, because Bytes is reference-counted. Still, if there are 100+ segments, all those reference counts need to be updated, and that is what is being changed here. And I didn't want to change SegmentedBytes. But yes, maybe there are better optimizations to get rid of the segments in the first place.

PR #197, opened 2025-11-30
performance optimizations: Signing key caching, stat_object to use HEAD

HJLebbink:
This PR implements performance optimizations across the MinIO Rust SDK, focusing on reducing latency and improving throughput for high-performance use cases like DataFusion/ObjectStore integration. Key improvements include signing key caching, HTTP/2 optimization, region lookup bypass, and fast-path APIs.

Key Changes:

Performance optimizations: Signing key caching (4 HMAC ops saved per request), HTTP/2 adaptive window, optimized query/header string building, and a fast-path GET API
Bug fixes: Corrected multipart copy logic, changed stat_object to use HEAD method, fixed region handling in tests
API enhancements: Added get_object_fast(), into_boxed_stream(), and into_bytes() methods for high-performance scenarios

### DIFF
--- a/src/s3/response/get_object.rs
+++ b/src/s3/response/get_object.rs
@@ -23,6 +23,13 @@ use bytes::Bytes;
 use futures_util::TryStreamExt;
 use http::HeaderMap;
 use std::mem;
+use std::pin::Pin;
+
+/// Type alias for a boxed byte stream with size, used by [`GetObjectResponse::into_boxed_stream`].
+pub type BoxedByteStream = (
+    Pin<Box<dyn futures_util::Stream<Item = std::io::Result<Bytes>> + Send>>,
+    u64,
+);
 
 pub struct GetObjectResponse {
     request: S3Request,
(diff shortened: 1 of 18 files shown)


mjs:
I just found this after discovering that stat_object uses GET instead of HEAD (a nasty bug, especially for large objects!). I'm looking forward to this landing.

src/s3/response/get_object.rs
```diff
@@ -47,6 +54,29 @@ impl GetObjectResponse {
...
+    ///
+    /// This is more efficient than `content().to_segmented_bytes().await` for
+    /// scenarios requiring minimal overhead, as it uses reqwest's native bytes
+    /// collection without intermediate wrappers.
+    pub async fn into_bytes(self) -> Result<Bytes, Error> {
+        self.resp
+            .bytes()
+            .await
+            .map_err(|e| ValidationErr::HttpError(e).into())
+    }
```
twuebi, line 79: This'll stream the whole request into memory, just something to keep in mind.

src/s3/signer.rs
```diff
@@ -25,8 +34,105 @@ use hyper::http::Method;
...
+///
+/// # Concurrency
+///
+/// Uses RwLock to allow concurrent reads (cache hits) while only blocking for writes.
+/// Uses Arc for zero-copy sharing of the signing key across threads.
+///
+/// This cache is stored per-client (in `SharedClientItems`) rather than globally
+/// to support multiple clients with different credentials in the same process.
+#[derive(Debug, Clone)]
+pub(crate) struct SigningKeyCache {
```
twuebi, line 70: Isn't this more of a precomputation than a Cache?

src/s3/response/get_object.rs
HJLebbink: I did not check by looking at memory consumption, but I think the alternative `content().to_segmented_bytes().to_bytes()` copies the payload to memory twice (once for the to_segmented_bytes, and once for the to_bytes, but I can be mistaken). I will update inline doc.

src/s3/signer.rs
HJLebbink: Yes, makes sense. lets update that.

src/s3/signer.rs
```diff
@@ -25,6 +34,98 @@ use hyper::http::Method;
 use ring::hmac;
 #[cfg(not(feature = "ring"))]
 use sha2::Sha256;
+use std::sync::{Arc, RwLock};
+
+/// Memoized precomputation of AWS Signature V4 signing keys.
+///
+/// Computing a signing key requires 4 HMAC-SHA256 operations. Since the key only
+/// changes when date, region, or service changes, we memoize the result to avoid
+/// redundant computation on subsequent requests.
```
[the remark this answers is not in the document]
HJLebbink: @twuebi What do you think, AI do not seem to agree. Is storing the hash of the signing key: a cache or a memoization?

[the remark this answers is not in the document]
twuebi: no strong opinions here, maybe my understanding of cache is also a bit too narrow, not a hill i'd be willing to die on 😄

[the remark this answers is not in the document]
HJLebbink: I'll use `cache` to please copilot: don't want to keep telling, "its ok".

PR #209, opened 2026-04-22
feat: use 64 MiB default part size for multipart uploads

HJLebbink:
- Increase DEFAULT_PART_SIZE from 16 MiB to 64 MiB for improved throughput
  - Update test expectations to reflect new default
  - Document part_size field with default value and auto-scaling behavior


klauspost (approved):
Sure. Was gonna suggest 32MB on your last PR, but thought it was getting into bike-shedding.  16, 32, 64 are all in the "fine" range.

PR #213, opened 2026-05-09
fix: x-amz-* starts_with conditions in add_starts_with_condition

aychun:
- fix `x-amz-meta-*` user metadata being rejected
- fix accepting any `x-amz-*` headers that are not `x-amz-meta-*`


HJLebbink:
@aychun Thanx for reporting this!

HJLebbink (approved):
LGTM

PR #217, opened 2026-06-08
feat: port pending minio-go changes + RenameObject, Inventory, QoS, Outposts, LDAP STS

harshavardhana:
## Summary

Brings `minio-rs` up to parity with recent `minio-go` changes, plus the MinIO/AIStor **RenameObject** extension. The work was scoped by diffing the recent `minio-go` commit window (`#2123` → `#2233`) against this repo and filling every gap, then validating the AIStor-only features against a live AIStor server.

## What's included

**Fixes / additions to existing APIs**
- Fix `DeleteObjectsResponse::result()` panic on an HTTP 200 response carrying a root `<Error>` body; add opt-in `expect_200_ok_with_error` detection in the execute path, enabled for `DeleteObjects` and `CompleteMultipartUpload` (go #2139).
- AppendObject post-call checks: error on missing `x-amz-object-size` and on object-size mismatch (go #2178).
- `x-minio-source-time` (`X-Minio-Source-Mtime`) override for `last_modified()` on GetObject/StatObject; add `GetObjectResponse::last_modified()` (go #2219).
- `content_encoding()` accessor on GetObject/StatObject (go #2208).
- Honor `StartAfter` when listing object versions via key-marker (go #2212).
- `ChecksumAlgorithm` on list response entries (go #2211).
- Omit empty lifecycle rule `<Filter>` (go #2177).
- Increase `MAX_OBJECT_SIZE` to `MAX_PART_SIZE * MAX_MULTIPART_COUNT` (~48.83 TiB) (go #2190).
- Conditional PUT (`If-Match`/`If-None-Match`) on PutObjectContent and CompleteMultipartUpload (go #2153).
- New checksum algorithms MD5, SHA512, XXHash64, XXHash3, XXHash128 across the enum, `compute_checksum`, streaming/aws-chunked hashers, and response parsing (go #2221, #2188).
- A public service-type-parameterized signer (`sign_v4_with_service_type`) (go #2214).
- ComposeObject `part_size` honoring `CopyDestOptions.PartSize` (go #2190).

**Net-new APIs**
- **GetObjectAttributes** with checksum & type fields and object parts (go #2217).
- **UpdateObjectEncryption** (`?encryption` PUT, version id from header) (go #2204, #2215).
- **GCS bulk delete** fallback: `BaseUrl::is_google_endpoint` + serialized single-DELETE path (go #2185).
- **QoS APIs** (Get/Set/Metrics), MinIO extension, YAML config bodies (go #2148).
- **Inventory APIs** (Generate/Put/Get/Delete/List/JobStatus) (go #2147..#2172).
- **LDAP STS** credentials provider (`AssumeRoleWithLDAPIdentity`) with `ConfigName` (go #2173).
- **RenameObject** — MinIO/AIStor single-request, same-bucket server-side rename (`PUT` + `x-amz-rename-source`), with source/destination conditional headers and directory-parity / prefix-nesting validation.

**Infrastructure**
- **S3 Outposts**: endpoint detection, region extraction, and `s3-outposts` service-type signing across header / streaming / presign paths.

## Validation

- 439 lib tests + 86 doc tests pass; `clippy --all-targets --all-features` clean; full build clean.
- The new AIStor-only features were exercised against a live AIStor server: `rename_object`, `get_object_attributes`, `bucket_qos`, `inventory`, and `conditional_put` pass end-to-end; the `update_object_encryption` request path is validated (its success path is KMS-key-gated).
- New integration tests live under `tests/s3/` and gate AIStor-only cases on `MINIO_AISTOR` so they no-op under the community-server CI. A new `aistor.yml` workflow runs them against a licensed AIStor server, sourcing the license from an `AISTOR_LICENSE` secret and self-guarding when the secret is absent.

## Caveats

- **LDAP STS is partial**: the `Provider::fetch()` trait is synchronous and infallible, so the provider primes its cache via an async `refresh()` rather than refetching transparently on expiry. Making it fully transparent needs async-ifying the `Provider` trait across all call sites — a separate follow-up.
- **UpdateObjectEncryption** success path requires a KMS-backed deployment and a valid key (`UPDATE_OBJECT_ENCRYPTION_KMS_KEY`).
- `rename_object` builds `x-amz-rename-source` raw (un-encoded object key), matching the existing `CopyObject` `x-amz-copy-source` convention.
- The AIStor CI workflow defaults to the `dl.min.io` server binary; if the AIStor feature set ships as a distinct binary, set `MINIO_SERVER_URL` in `tests/start-aistor-server.sh`.

## Summary by CodeRabbit

* **New Features**
  * Object attributes API, server-side rename, in-place encryption rotation, bucket QoS, and inventory management
  * LDAP-backed STS provider, Outposts-aware signing, Google-delete fallback, expanded checksum algorithms, conditional ETag uploads, append-size validation

* **Documentation**
  * Added a pending-from-minio-go compatibility & porting checklist

* **Tests**
  * New/expanded unit and integration tests covering QoS, inventory, rename, checksums, conditional puts, and encryption

* **Dependencies**
  * Added YAML and XXHash support libraries


harshavardhana:
Thanks @coderabbitai — addressed the substantive findings in 6aefc3b:

**Fixed**
- **delete_objects (GCS single-DELETE)**: XML-escape object keys / version ids / error code+message in the synthesized `<DeleteResult>` (keys can contain `&`/`<`, which would otherwise produce invalid XML that fails to re-parse).
- **get_object_attributes**: invalid `version_id` now propagates as `ValidationErr` instead of panicking during request build.
- **LDAP STS**: added a 60s timeout to the STS HTTP request.
- **inventory list-all**: break the pagination loop if the server repeats a continuation token (infinite-loop guard).
- **update_object_encryption**: reject whitespace-only `kms_key_arn`.
- **tests**: `rename_object` now strictly asserts the rename (dst has data, src gone) — no capability fallback; `update_object_encryption` tolerance tightened to KMS/encryption errors only; `get_object_attributes` asserts a non-empty ETag.

**Intentionally not changed**
- **AIStor license token in `tests/start-server.sh`**: this is the public FREE-tier dev license and is deliberately committed so CI can run AIStor-only APIs without a secret.
- **`edge` binary without pinning**: intentional — RenameObject and other features only exist in the edge channel right now; pinning to a release would drop them.
- **`get_object_attributes` numeric XML coercion to 0** and **delete root-element name validation**: left as-is (server responses are trusted; low real-world risk).

All verified against the edge AIStor server (`EDGE.2026-06-06`).

PR #225, opened 2026-06-19
feat: recursive prefix rename (x-amz-rename-recursive)

harshavardhana:
## What

Add a native `recursive` flag to the `RenameObject` builder. When set it emits the `x-amz-rename-recursive: true` header, turning the request into the MinIO/AIStor **recursive prefix rename** (`RenamePrefix`): every object under the source directory prefix is moved into the destination prefix in one server-side request.

Source and destination must be directory prefixes (trailing `/`), already validated by `validate_rename` (which also rejects directory/object-suffix mismatch and prefix nesting).

## Why

minfs (the POSIX FUSE mount over AIStor) needs server-side directory rename — re-keying every object client-side is O(n) round-trips. The server-side `RenamePrefix` collapses it into one journaled, all-or-nothing request. This is the client-side half.

## Notes on the slow-rename / rollback wire contract

A recursive rename of many or large objects can take time, so AIStor keeps the connection alive with whitespace (the `CompleteMultipartUpload` pattern): it sends `200 OK` + the `<?xml?>` header, streams spaces while it works, then writes the final body — which on a rollback is a root `<Error>` element **under the 200**. The builder already opts into `expect_200_ok_with_error(true)`, so `execute_internal` reads the full body and surfaces that `<Error>` as a server error. Added unit tests for the exact `<?xml?>` + keep-alive-whitespace + `<Error>` / `<RenamePrefixResult>` layouts.

## Changes

- `RenameObject` builder: `recursive` field (+ tuple-alias slot) emitting `x-amz-rename-recursive`.
- `X_AMZ_RENAME_RECURSIVE` header constant.
- `rename_prefix` integration test (puts a small tree, renames recursively, asserts every child moved and the source prefix is empty).
- Test harness (`tests/start-server.sh`) now runs the `registry.min.dev/aistor/minio:edge` docker image (carries `RenamePrefix`) instead of downloading the binary.

## Test plan

- `cargo test --lib rename` — builder + 200-with-error parse unit tests pass.
- `rename_prefix` integration test runs against the `:edge` server.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
  * Added support for recursive S3 rename operations on object prefixes.

* **Bug Fixes**
  * Adjusted rename request headers to use key-only source formatting for rename operations.

* **Tests**
  * Added an integration test covering recursive prefix renames.
  * Added unit tests to verify correct handling of “200 OK with error body” XML parsing edge cases.

* **Chores / Infrastructure**
  * Updated the MinIO test server setup to run via Docker with readiness polling.

### DIFF
--- a/src/s3/builders/rename_object.rs
+++ b/src/s3/builders/rename_object.rs
@@ -122,14 +122,15 @@ fn conditional_etag_value(etag: &str) -> String {
     }
 }
 
-/// Builds the `x-amz-rename-source` header value as `/{bucket}/{object}`, percent-encoding
-/// the object key (preserving `/`) the same way minio-go encodes `x-amz-copy-source`.
-/// The server decodes it via `url.Parse().Path`, so the key must be encoded here.
-fn rename_source_value(bucket: &str, src_object: &str) -> String {
+/// Builds the `x-amz-rename-source` header value as `/{object}` — the source
+/// **key only**, with no bucket qualifier. AIStor addresses the rename source by
+/// key within the request bucket (same-bucket rename); the whole header value is
+/// the key and any slashes are key prefixes. The server decodes it via
+/// `url.Parse().Path` then `trimLeadingSlash`, so the key is percent-encoded here
+/// (preserving `/`) and prefixed with a single leading slash.
+fn rename_source_value(src_object: &str) -> String {
     let encoded = urlencode_object_key(src_object);
-    let mut value = String::with_capacity(bucket.len() + encoded.len() + 2);
-    value.push('/');
-    value.push_str(bucket);
+    let mut value = String::with_capacity(encoded.len() + 1);
     value.push('/');
     value.push_str(&encoded);
     value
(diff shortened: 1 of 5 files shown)


harshavardhana:
Pushed fixes for CI + review feedback:

- **`x-amz-rename-source` was bucket-qualified.** The builder emitted `/{bucket}/{key}`, but current AIStor reads the header as the **key only** (`url.Parse().Path` → `trimLeadingSlash`, no bucket stripping), so single renames 404'd the source and recursive prefix renames silently moved nothing (`Moved=0`, HTTP 200). Now emits `/{key}`. Verified end-to-end against `registry.min.dev/aistor/minio:edge`: both single and recursive prefix renames succeed and move the objects; the old `/{bucket}/{key}` form 404s. This also fixes the pre-existing `rename_object` integration test.
- **rustfmt** on the `rename_prefix` test.
- **CodeRabbit (start-server.sh):** wrapped the `MINIO_LICENSE` + `docker run` in `set +x`/`set -x` so the credential isn't echoed in CI logs; replaced `sleep 10` with a bounded `/minio/health/ready` probe (`--cacert tests/public.crt`).

PR #230, opened 2026-07-10
feat: add object annotation API (put/get/list/delete)

harshavardhana:
## What

Adds the object-annotation subresource operations to the S3 client, mirroring the existing `minio-go` implementation and the server contract:

- `put_object_annotation(bucket, object, name, payload)` — `PUT /{object}?annotation&annotationName=N`, returns the annotation ETag
- `get_object_annotation(bucket, object, name)` — `GET /{object}?annotation&annotationName=N`, body is the annotation payload
- `list_object_annotations(bucket, object)` — `GET /{object}?annotation`, `annotations()` yields the parsed entries
- `delete_object_annotation(bucket, object, name)` — `DELETE /{object}?annotation&annotationName=N`

## Details

An annotation is a named payload (1 byte–1 MiB of UTF-8 text) attached to a specific object version, independent of the object's data; up to 1000 may be attached per version.

- Names validated client-side (non-empty, ≤512 bytes) and payload size enforced (1 B–1 MiB) in `to_s3request`, before the request is sent.
- `put`/`delete` accept an optional `x-amz-object-if-match` conditional header; all ops support `version_id`.
- Follows the existing tagging operations' builder/client/response 3-file layout; signing/subresource handling reuses the generic `S3Request` path unchanged.
- List XML (`<Annotation>` entries directly under the root — no `TagSet` wrapper) is parsed tolerantly: a bad `LastModified` degrades to `None` rather than failing the whole listing.

## Testing

- `cargo build`, `cargo clippy --lib --tests`, `cargo fmt --check` all clean (default features and `--no-default-features --features rustls-tls,default-crypto,http2`).
- Unit test covers List XML parsing (ETag quote-trimming + malformed-timestamp tolerance).
- End-to-end round-trip against a live server is not included here (requires a backend exposing the `?annotation` endpoint).

## Summary by CodeRabbit

* **New Features**
  * Added S3/MinIO object annotation APIs to put, get, list, and delete annotations, including request builders with region and optional versioning, plus extra headers/query parameters.
  * Added typed response types, including XML parsing for listing annotations and accessors for annotation payload/metadata.
* **Validation**
  * Added dedicated validation for annotation names (non-empty, max length) and payload sizes (1 byte–1 MiB), with new error variants.
  * Supported conditional requests via the `x-amz-object-if-match` header for put/delete.
* **Bug Fixes**
  * Listing degrades gracefully when individual entries have invalid timestamp data.
* **Tests**
  * Updated the integration test server container image reference.


harshavardhana:
## CodeRabbit Autofix Applied

Addressed the review feedback in one follow-up commit:

- **Typed `AnnotationName` wrapper** — added a `TryFrom`-based newtype (validated non-empty, ≤512 bytes) per the repo's typed-parameter convention, replacing runtime `String` validation in the put/get/delete builders. Public call sites stay `&str`-ergonomic (validation happens at the boundary).
- **Same-file unit tests** — for the builders' `to_s3request` construction (subresource + `annotationName`/`versionId` query params, `x-amz-object-if-match` header), the `AnnotationName` and payload-size validators, and a malformed-XML case for the list parser.
- **`#[must_use]`** on `GetObjectAnnotationResponse::into_payload`.

Deferred the client-seed-method test (thin wrapper; behavior is already covered by the builder `to_s3request` tests, consistent with the other client modules).

Verified: `cargo fmt --check`, `cargo clippy --lib --tests`, and the annotation test suite (19 tests) all clean.

harshavardhana:
## CodeRabbit Autofix — round 2

Follow-up commit `00cac01` completes the review coverage:

- **Expanded builder request-construction tests** (put/get/list/delete) — now assert the HTTP method, `versionId` propagation, the exact `x-amz-object-if-match` value (not just presence), attached body, oversized-payload rejection, and absence of conditional headers on GET.
- **Client-method test** for `delete_object_annotation` (wiring + empty-name rejection).
- **`S3Request::method` and `body` → `pub(crate)`** — consistent with the already-`pub(crate)` `query_params`/`headers`/`object`, so builder tests can assert the verb and body without a live server.

Annotation test suite now 22 tests; `cargo fmt --check`, `cargo clippy --lib --tests` clean.

### One suggestion intentionally not applied

The `TryFrom<&String>` nitpick on `AnnotationName` is left as-is: every sibling typed wrapper in `typed_parameters.rs` (`ObjectKey`, `VersionId`, `BucketName`) provides the same `TryFrom<&String>` impl, so removing it from only `AnnotationName` would be an inconsistency. Happy to drop it repo-wide in a separate change if that's the preferred direction.

PR #231, opened 2026-07-17
feat(signer): expose public SigV4 request-signing helpers

harshavardhana:
## Summary

Exposes public SigV4 request-signing helpers so external crates can sign
arbitrary requests to non-object S3 APIs served on the same endpoint (the AIStor
Memory API under `/_mem/v1` and the Tables/Iceberg REST catalog under
`/_iceberg/v1`). Previously the `sign_v4_*` functions were `pub(crate)`, so a
consumer had to re-implement SigV4 to talk to these APIs.

## What's added

- `sign_v4_s3_request(...)` — self-contained: owns its signing-key cache and adds
  the mandatory `host` / `x-amz-date` / `x-amz-content-sha256` (and
  `x-amz-security-token` when a session token is supplied) headers before signing,
  returning the complete header set. Signs as service `s3`.
- `sign_v4_service_request(service, ...)` — same, parameterized by service type
  (`s3tables` for the Iceberg REST catalog, `sts`, etc.); the service becomes part
  of the credential scope. `sign_v4_s3_request` now delegates to it with `s3`.

Both build on the existing internal `sign_v4_*` machinery — no new dependencies,
one SigV4 implementation.

## Motivation

`aimem` (the agent-native filesystem) needs to call the AIStor Memory API and
Tables API on the same endpoint as object I/O. These are not object APIs, so the
SDK's per-client signing path doesn't apply; a small public signing entrypoint
lets the client sign those REST requests with the same, proven SigV4 code rather
than a second implementation. Verified end-to-end against a live AIStor server
(cortex CRUD/secrets signed as `s3`; Iceberg catalog signed as `s3tables`).

## Summary by CodeRabbit

* **New Features**
  * Added support for signing non-object S3-compatible API requests using AWS Signature Version 4.
  * Introduced a reusable request-signing API that accepts consolidated signing options (service scope, method/host/path, query params, headers, credentials, timestamp, payload hash).
  * Automatically injects required authentication headers and includes session-token authentication when provided.
* **Tests**
  * Expanded unit tests to validate header injection/signed-headers behavior, service scoping differences, and deterministic reuse of signing-key caching.

### DIFF
--- a/src/s3/signer.rs
+++ b/src/s3/signer.rs
@@ -381,6 +381,41 @@ pub fn sign_v4_s3_request(
     payload_sha256: &str,
     date: UtcTime,
 ) -> Multimap {
+    sign_v4_service_request(
+        "s3",
+        method,
+        host,
+        path,
+        region,
+        query_params,
+        extra_headers,
+        access_key,
+        secret_key,
+        session_token,
+        payload_sha256,
+        date,
+    )
+}
+
+/// Like [`sign_v4_s3_request`] but signs for an arbitrary `service` type (e.g.
+/// `"s3tables"` for the Iceberg REST catalog, or `"sts"` for STS actions). The
+/// `service` becomes part of the credential scope
+/// (`.../<region>/<service>/aws4_request`).
+#[allow(clippy::too_many_arguments)]
+pub fn sign_v4_service_request(
+    service: &str,
+    method: &Method,
+    host: &str,
+    path: &str,
+    region: &str,
+    query_params: &Multimap,
+    extra_headers: &Multimap,
+    access_key: &str,
+    secret_key: &str,
+    session_token: Option<&str>,
+    payload_sha256: &str,
+    date: UtcTime,
+) -> Multimap {
     let mut headers = extra_headers.clone();
     headers.add(HOST, host);
     headers.add(X_AMZ_DATE, to_amz_date(date));
@@ -391,8 +426,9 @@ pub fn sign_v4_s3_request(
 
     let cache = RwLock::new(SigningKeyCache::new());
     let region = Region::new(region.to_owned()).unwrap_or_default();
-    sign_v4_s3(
+    sign_v4_with_service_type(
         &cache,
+        service,
         method,
         path,
         &region,
--- a/src/s3/signer.rs
+++ b/src/s3/signer.rs
@@ -347,99 +347,98 @@ pub(crate) fn sign_v4_s3(
     )
 }
 
-/// Signs an arbitrary HTTP request with AWS Signature V4 as service `"s3"` and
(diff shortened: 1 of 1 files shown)


harshavardhana:
## CodeRabbit feedback addressed

Reworked the public signing helpers in `src/s3/signer.rs` in one commit, addressing all four review threads:

- **Stateful cache** — replaced the two free functions (which created a throwaway `SigningKeyCache` per call) with a `RequestSigner` struct that owns a persistent cache reused across `sign()` calls. Removed the inaccurate "owns its signing-key cache" doc claim.
- **Parameter struct + `#[must_use]`** — consolidated the long argument list into `SignRequestOptions`; dropped `#[allow(clippy::too_many_arguments)]`; added `#[must_use]` to `new()` and `sign()`.
- **Owned headers** — `SignRequestOptions::extra_headers` is taken by value; the signer mutates and returns it, no internal clone.
- **Unit tests** — added same-file tests covering mandatory-header addition, extra-header preservation, per-service credential scope (`s3` vs `s3tables`), the session-token header, and cache-reuse determinism.

`cargo fmt`, `clippy --all-targets`, and doc tests are clean; the new unit tests pass. Verified end-to-end against a live AIStor server (Memory API signed as `s3`, Iceberg catalog as `s3tables`).

**Commit:** 8664116

harshavardhana:
## Round-2 CodeRabbit feedback addressed

Reworked `RequestSigner` in `src/s3/signer.rs` (commit 16224b4), addressing all four new threads:

- **Credential-scoped cache** — the credential (access/secret/session) is now bound to `RequestSigner::new(...)` rather than passed per `sign()` call. The signing-key cache is therefore per-credential by construction, so a rotated secret can never reuse a key derived from the previous one (rotation → new signer, matching the SDK's per-client cache design). Added a redacting `Debug` impl.
- **Typed region** — `SignRequestOptions::region` is now `&Region` (caller-validated), removing the silent `Region::default()` fallback that could produce a valid-looking signature for the wrong credential scope.
- **Reserved-header safety** — `sign()` strips any caller-supplied `host` / `x-amz-date` / `x-amz-content-sha256` / `x-amz-security-token` (case-insensitively) before adding its own, so those headers can never be duplicated and canonicalization matches transmission.
- **Cache-reuse test** — replaced the determinism-only assertion with an `Arc::ptr_eq` check on the cached signing key, so a regression to per-call derivation fails the test. Also added tests for credential binding (different secret → different signature) and reserved-header stripping.

`cargo fmt`, `clippy --all-targets`, and the 7 signer unit tests are clean. Re-verified end-to-end against a live AIStor server (Memory API `s3`, Iceberg catalog `s3tables`, and STS-scoped credentials).

**Commit:** 16224b4

PR #233, opened 2026-08-04
rdma: reject buffers over the 4 GiB cuObject registration limit

harshavardhana:
## What

A single cuObject registration (`cuMemObjGetDescriptor`) can pin at most **4 GiB**. The RDMA methods registered the whole caller buffer unconditionally, so an oversized buffer failed with the opaque `RegistrationFailed`. This rejects it up front with an actionable error.

- Add `CUOBJ_MAX_MEMORY_REG_SIZE` (4 GiB) and `RdmaError::BufferTooLarge { size, max }`.
- Guard all three registration sites — `rdma_put_object`, `rdma_get_object`, `rdma_upload_part`. `rdma_put_object_multipart` delegates to `rdma_upload_part`, so per-part sizes are covered there too. This also bounds the RDMA multipart part size, which otherwise bypassed the 5 GiB builder validation.
- Unit-test the inclusive boundary (≤ limit ok, > limit → `BufferTooLarge`).

## Why fail-fast (not a silent fallback)

These methods are RDMA-only by design — the existing error variants already instruct callers to fall back to HTTP `PutObject`/`GetObject`. So a clear, distinct error is the right behavior; sizing the buffer stays the caller's responsibility. (This mirrors the equivalent fix landed in minio-cpp, adapted to minio-rs's RDMA-only method design.)

## Testing

`cargo check/test/clippy --features rdma` — builds the cuObject shim + links ibverbs/rdmacm/numa; the new `ensure_registrable_boundary` unit test passes; clippy clean.

## Follow-up (not in this PR)

`rdma_get_object` has no object-offset parameter, so it can't yet stream a > 4 GiB object as a sequence of ≤ 4 GiB ranged reads (the minio-cpp ranged-GET counterpart). Worth a separate change if that use case is needed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to prevent RDMA uploads, downloads, and multipart uploads from exceeding the supported 4 GiB buffer limit.
  * Oversized buffers now return a clear error before registration, while buffers exactly at the limit remain supported.
  * Improved handling and diagnostics when encryption-related server features, licensing, or KMS provisioning are unavailable.

### DIFF
--- a/src/s3/rdma/client.rs
+++ b/src/s3/rdma/client.rs
@@ -32,7 +32,11 @@ pub struct RdmaResponse {
 }
 
 /// Errors specific to the RDMA fast path.
+///
+/// `#[non_exhaustive]` so future variants (e.g. `BufferTooLarge`) can be added
+/// without breaking downstream exhaustive `match` expressions.
 #[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
 pub enum RdmaError {
     #[error("RDMA not available: cuObjClient not connected (no cuObjServer reachable)")]
     NotConnected,
--- a/src/s3/rdma/client.rs
+++ b/src/s3/rdma/client.rs
@@ -33,8 +33,8 @@ pub struct RdmaResponse {
 
 /// Errors specific to the RDMA fast path.
 ///
-/// `#[non_exhaustive]` so future variants (e.g. `BufferTooLarge`) can be added
-/// without breaking downstream exhaustive `match` expressions.
+/// `#[non_exhaustive]`: new variants can be added in future releases without
+/// requiring downstream `match` expressions to change for each one.
 #[derive(Debug, thiserror::Error)]
 #[non_exhaustive]
 pub enum RdmaError {
--- a/src/s3/rdma/client.rs
+++ b/src/s3/rdma/client.rs
@@ -20,7 +20,7 @@ use crate::s3::types::{BucketName, ETag, ObjectKey, PartInfo, Region, S3Api, Upl
 use crate::s3::utils::ChecksumAlgorithm;
 
 use super::buffer::RdmaBuffer;
-use super::cuobj::{ScopedRegistration, shared};
+use super::cuobj::{CUOBJ_MAX_MEMORY_REG_SIZE, ScopedRegistration, shared};
 use super::protocol::{RdmaOutcome, S3RdmaClientCtx, rdma_get_with_retry, rdma_put_with_retry};
 
 /// Successful RDMA transfer result.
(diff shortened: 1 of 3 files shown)


harshavardhana:
## CodeRabbit Autofix — Applied

Addressed 1 of 2 review items in `f9066c3`.

**Applied:**
- 🟡 *Preserve stable error matching for `RdmaError`* — marked the enum `#[non_exhaustive]` so adding `BufferTooLarge` (and future variants) is not a breaking change for downstream exhaustive `match`. — `src/s3/rdma/client.rs`

**Deferred (with reason):**
- 🔵 *Cover the three RDMA entry points with boundary tests* — CodeRabbit marks this "Heavy lift", and correctly so: the three methods require a **connected** cuObjClient (they return `NotConnected` before the size guard in any test environment), so entry-point boundary tests need a mockable registration/DI layer the crate doesn't have. The guard logic is already unit-tested (`ensure_registrable_boundary`, covering zero / exact-limit / over-limit) and the three call sites are one-line, trivially auditable. Better as a scoped refactor if entry-point mocking is introduced.

**Validated:** `cargo check / clippy / test --features rdma` — all pass.

harshavardhana:
## CodeRabbit Autofix — Applied

Addressed 1 of 2 open items in `8bbd10a`.

**Applied:**
- 🟡 *Correct the `RdmaError` documentation* — the `#[non_exhaustive]` doc no longer calls `BufferTooLarge` a "future" variant (it's current); it now states what `#[non_exhaustive]` provides. — `src/s3/rdma/client.rs`

**Deferred (with reason):**
- 🔵 *Cover the three RDMA entry points with boundary tests* — CodeRabbit marks it "Heavy lift"; the methods need a connected cuObjClient (they return `NotConnected` before the guard in a test env), so entry-point tests need a mockable registration/DI layer the crate lacks. The guard is already unit-tested (`ensure_registrable_boundary`) and the call sites are one-liners.

Added the `enhancement` label to satisfy the label check. The `test-*` job failure is unrelated: it's the `s3::update_object_encryption` integration test against the CI's (latest, unpinned) MinIO server, which the same run reports lacks features (`server does not support S3 Select`) — my rdma-feature-gated changes aren't compiled in that job (528 lib tests pass).

harshavardhana:
## CodeRabbit Autofix — Applied

Addressed 1 item in `ec430b4`.

**Applied:**
- 🟠 *Classify expected failures by structured server errors* — the `update_object_encryption` failure branch now matches `Error::S3Server(S3ServerError::S3Error(e))` and asserts on the controlled S3 **error code** (license/kms/notimplemented/methodnotallowed/notsupported), instead of substring-matching the rendered message. Client-side signing/parse/transport errors are a distinct arm that fails the test, so the "reached the server" claim is reliable and an SDK regression can't be masked. — `tests/s3/update_object_encryption.rs`

**Validated:** `cargo fmt` + `cargo check --tests` clean.

PR #234, opened 2026-08-07
rdma: send the descriptor as x-amz-rdma-token; drop the addr:size append

harshavardhana:
## What

`rdma_put`/`rdma_get` no longer append `:addr:size` to the RDMA token. The token from `cuMemObjGetRDMAToken` **is** the cuObject descriptor, whose leading fields already carry the buffer address and transfer size:
```
<addr:16>:<size:8>:<rkey:8>:<lid:4>:<qp:6>:<gid_present:1>:<gid:32>
```
Appending `:addr:size` merely duplicated fields 0 and 1, so `format_rdma_token` is removed, the descriptor is sent verbatim, and the unused `buf_addr` argument is dropped from `rdma_put`/`rdma_get`.

## Validation

Probed on a live mlx5 fabric (libcuobjclient 1.2.0), both GET and PUT:
```
token = "000073967e5ff000:00400000:01f2e973:0000:007a03:1:...ffff0f0f0fdf"
  field[0] == ptr  ? YES
  field[1] == size ? YES
```

`cargo check`, `cargo fmt --check`, `cargo clippy` all clean with `--features rdma`.

## Companion changes

Server reads addr/size from the descriptor's own fields (miniohq/aistor#6758, #6759). Also: minio-cpp#244, nixl ai-dynamo/nixl#1840. minio-go/minio-py inherit the fix through libminiocpp (no change needed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RDMA data transfers by preserving transfer tokens unchanged during PUT and GET operations.
  * Updated retry handling to use the correct token and transfer size information.

### DIFF
--- a/src/s3/rdma/protocol.rs
+++ b/src/s3/rdma/protocol.rs
@@ -173,21 +173,17 @@ fn http_client_for_token(token: &CStr) -> Arc<reqwest::Client> {
     }
 }
 
-fn format_rdma_token(token: &CStr, buf_addr: u64, size: u64) -> String {
-    let s = token.to_str().unwrap_or("");
-    format!("{s}:{buf_addr:016x}:{size:016x}")
-}
-
 /// Mirrors C++ `rdmaPut`: signs and issues the HTTP PUT control plane carrying
 /// the RDMA token, then parses the server reply.
 pub async fn rdma_put(
     client: &MinioClient,
     ctx: &mut S3RdmaClientCtx,
     token: &CStr,
-    buf_addr: u64,
     size: u64,
 ) -> isize {
-    let rdma_token = format_rdma_token(token, buf_addr, size);
+    // The token is the RDMA descriptor verbatim; its leading fields already
+    // carry the buffer address and transfer size, so it is sent as-is.
+    let rdma_token = token.to_str().unwrap_or("");
 
     let mut query_params = Multimap::default();
     if let Some(upload_id) = &ctx.upload_id {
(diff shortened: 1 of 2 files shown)


harshavardhana:
## CodeRabbit autofix — reviewed, no code change

One unresolved item: *"Add same-file coverage for verbatim RDMA token forwarding"* (🟡 Minor). **Deferred**, with reasoning:

The token is now forwarded verbatim — `let rdma_token = token.to_str().unwrap_or("")`, with no transformation — because this change's whole premise is that the cuObject descriptor already carries the buffer address and size in its own leading fields, so nothing is appended.

A faithful assertion that `rdma_put`/`rdma_get` emit `x-amz-rdma-token` unchanged would need a mock HTTP layer: there's no header-inspection seam (headers are built inline, then signed and sent), and the repo has no wiremock/httpmock harness (`mod tests` here is synchronous parser tests only). The sibling client changes — minio-cpp#244, ai-dynamo/nixl#1840 — make the identical verbatim-forwarding change without adding such tests, so I'm skipping to keep parity and avoid disproportionate test infra for a one-line forwarding. Happy to revisit if a mock-server harness is introduced.

harshavardhana:
## CodeRabbit autofix — applied

**Fixed 1 CodeRabbit item (🟠 Major):** *Add the required AWS API references.*
- `src/s3/rdma/protocol.rs` — `rdma_put` now documents both [PutObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html) and [UploadPart](https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html) (the `uploadId` path selects the multipart form); `rdma_get` documents [GetObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html). Matches the builder-pattern doc guideline. (`50ee2b2`)

**CI fix (separate):** the `test-current-thread`/`test-multi-thread` jobs were failing on `docker pull registry.k5.min.dev/aistor/minio:edge` (*context deadline exceeded* — fork-PR runners can't reach the private registry). `tests/start-server.sh` now falls back to the public `quay.io/minio/aistor/minio:edge-daily` mirror when the private pull fails. (`e8c035f`)

`cargo fmt --check` + `cargo check --features rdma` clean.

PR #235, opened 2026-08-15
feat(rdma): move the data path to libs3rdma

harshavardhana:
The RDMA path bound NVIDIA's `libcuobjclient` through a C++ shim. This moves it to `libs3rdma`, which speaks the same `x-amz-rdma-token` wire format over plain IBTA verbs.

## What the crate supports now that it did not before

- **Native InfiniBand**, not just RoCE.
- **Any HCA**, with no CUDA-adjacent client library on the host.
- **Building with no C++ toolchain.** `cuobj_shim.cc` / `.h` are deleted and the `cc` build-dependency goes with them. `build.rs` now points the linker at the vendored library and does nothing else — no `cc::Build`, no C++ compile step.

The link line is just `-ls3rdma`: the library resolves the RDMA stack itself at run time, so `libibverbs`/`librdmacm`/`libnuma` are no longer link-time dependencies and are dropped from CI.

## API changes

`CuObjClient` becomes `RdmaClient`, in a new `transport` module (`cuobj.rs` → `transport.rs`). Public re-exports change accordingly: `shared_cuobj_client` → `shared_rdma_client`, and `RdmaToken` is now exported.

`OpType` is removed. The token carries no direction — the server takes that from the S3 request, and the memory is registered for both READ and WRITE — so a GET token and a PUT token were byte-identical.

## Two behaviours changed, deliberately

- **`rdma_available()` reports a usable local RDMA device, not a reachable server.** DC is connectionless, so there is no session to probe the way `cuObjClient::isConnected` probed a cuObjServer. A server that will not serve RDMA still declines per request with `x-amz-rdma-reply: 501` and the caller falls back to HTTP — the fallback behaviour is unchanged, only the meaning of this predicate is.
- **The registration ceiling is `u32::MAX`, not 4 GiB** (`RDMA_MAX_MEMORY_REG_SIZE`): the descriptor carries the window size in a 32-bit field. A buffer of exactly 4 GiB would previously pass this guard and then be declined by the transport.

## Requirements

The client mints DC tokens, so it needs a DC-capable HCA. On a device without DC, `rdma_available()` is false and every transfer takes the ordinary HTTP path.

`libs3rdma` is vendored for **both** `x86_64` and `aarch64` under `vendor/s3rdma/`, built by `build-libs.sh` in the s3rdma repo.

## Verification

`cargo build --all-targets --features rdma`, `cargo clippy --all-targets --features rdma -- -D warnings`, `cargo fmt --check` and the 538 lib tests all pass.

The underlying client FFI is verified on a ConnectX HCA: token layout, the IPv4-mapped GID tail that `parse_client_nic_from_token` depends on, offset windows, oversize rejection, and concurrent register/mint/deregister.

Requires miniohq/s3rdma#15.

Not covered here: a live PUT/GET round trip against a server, which needs an endpoint.

## Summary by CodeRabbit

* **New Features**
  * Added RDMA transport support for Linux x86_64 and aarch64 systems.
  * Added readiness checks, memory registration, token management, memory-type detection, and NIC rail health reporting.
  * Supports RoCE and InfiniBand, with graceful fallback when RDMA is unavailable.
  * Added an example for inspecting RDMA rail availability and distribution.

* **Refactor**
  * Replaced the previous NVIDIA-specific RDMA integration with a generic transport implementation.
  * Removed obsolete GPU Direct Storage and legacy RDMA components.


harshavardhana:
Re-vendored at **libs3rdma 0.2.0** and brought the crate to parity with minio-cpp on dual-NIC handling.

## What comes from the library

Multi-rail itself is inside libs3rdma: every device with an ACTIVE port is opened, a pinned buffer is registered on **all** of them, and each token names one rail chosen round-robin. The crate got that for free by re-vendoring.

Verified on a dual-rail ConnectX host with a new `rdma_rails` example:

```
rails: 2 total, 2 healthy
  token 0 gid=...ffff0f0f0fdf     <- 15.15.15.223
  token 1 gid=...ffff0f0f10f1     <- 15.15.16.241
  token 2 gid=...ffff0f0f0fdf
  ...
distinct rails across 6 mints: 2
```

## What the SDK half was missing

**A failed transfer never told the library which rail carried it.** The retry re-mints and happens to land on the other rail, so a single request recovers — but the *next* request round-robins straight back onto the dead rail. Every request kept paying a failed attempt against a NIC that was known bad.

`rdma_put_with_retry` / `rdma_get_with_retry` now report it, which takes the rail out of rotation. Two deliberate choices:

- **A 501 is excluded.** The server declining RDMA says nothing about the rail, and marking one for that would sideline healthy hardware over a server-side policy.
- **A server-side fault marks every rail in turn, and that is safe.** The library clears all marks once no rail is left usable, so a fault that was never the fabric's heals itself rather than stranding the client.

**The reason RDMA is unavailable was discarded.** `s3rdma_client_init` reports it into a caller-supplied buffer and the crate passed NULL — so a host with no hardware and a host with a *misconfigured* HCA both silently fell back to HTTP, indistinguishable. Now kept and exposed as `rdma_unavailable_reason()`.

**Rail counts are exposed** — `rdma_nic_count()` and `rdma_healthy_nic_count()`. A healthy count below the total is a client serving on what is left, which was otherwise invisible.

## Note on the NIC-pinning path

This crate was already ahead of minio-cpp there: `parse_client_nic_from_token` plus a `reqwest` client cached **per NIC IP**, so each rail keeps its own connection pool. minio-cpp sets `CURLOPT_INTERFACE` per request instead. Nothing needed changing, but it is worth knowing the two differ.

`cargo build --all-targets --features rdma`, `clippy -- -D warnings`, `fmt --check` and the 538 lib tests all pass.

harshavardhana:
@coderabbitai review

PR #240, opened 2026-09-14
perf(aws-chunked): stream chunk payloads by reference instead of copying

HJLebbink:
## What changed

A streaming upload copied every payload byte three times before it reached the
wire: into the rechunk buffer, out of that buffer into a fresh `Vec`, and once
more into the joined chunk frame. On the common path no payload byte is copied
now.

- `RechunkingStream` slices each chunk out of the incoming `Bytes`, which shares
  the input buffer by reference count. An input piece shorter than the chunk size
  is accumulated in a `BytesMut`, topped up to exactly one chunk and no further,
  so the remainder of a large piece is still served by reference.
- `AwsChunkedEncoder` and `SignedAwsChunkedEncoder` emit a chunk as three pieces,
  the header, the payload and the closing CRLF, held in a `VecDeque` between
  polls. Joining the three into one buffer per chunk is what forced the payload
  copy.

The bytes on the wire are unchanged, and the chunk signature covers the payload
alone, which the split does not alter.

## Measured effect

256 MiB body, 64 KiB chunks, CRC32C, release build, no network, single-threaded
runtime, Intel Xeon w5-2455X. Seven runs per case after a warm-up, fastest run
reported.

| Input piece size | Before | After | Speedup |
| --- | --- | --- | --- |
| 8 KiB | 34.6 ms | 23.2 ms | 1.49x |
| 64 KiB | 34.2 ms | 16.3 ms | 2.10x |
| 256 KiB | 40.5 ms | 16.3 ms | 2.49x |
| 1 MiB | 70.6 ms | 16.4 ms | 4.32x |

Allocation traffic for the same body falls from 512 MiB to 0.19 MiB, and
allocation count from 12293 to 8198. The remaining allocations are the per-chunk
header strings, one per 64 KiB chunk.

Two caveats. Larger input pieces gain more because the old rechunker also shifted
the rest of the buffered piece down on every chunk it drained, so part of the
4.32x is that effect rather than the copy removal. A real upload adds network and
TLS, which usually dominate, so the end-to-end gain is smaller than the table.

## Tests

Two tests guard the invariants the wire depends on.

- The encoded stream is exactly as long as `calculate_encoded_length` declared,
  over eight body sizes crossed with three input segmentations. The encoder sends
  that length as `Content-Length` before the body, so a mismatch fails the upload
  rather than the test suite.
- The rechunker returns the original bytes in order, as full chunks followed by at
  most one short tail, for any input segmentation.

## Summary by CodeRabbit

* **Performance**
  * Improved streaming efficiency for chunked uploads, reducing unnecessary buffering and data copying.
  * Large or segmented request bodies are now processed more smoothly.

* **Bug Fixes**
  * Improved consistency of chunk boundaries and encoded stream lengths across varying body sizes and segmentations.
  * Added validation to reject invalid zero-sized chunk configurations.


HJLebbink:
@coderabbitai what needs to change for you to approve?

PR #244, opened 2026-09-21
fix(types): validate a bucket name once, against the strict rules

HJLebbink:
## Description

`BucketName::new` applied the relaxed MinIO rules, and 28 sites across 17 request
builders applied the strict S3 rules a second time inside `to_s3request`. The two
rule sets differ: relaxed allows uppercase, `_` and `:`, and allows the names AWS
reserves (`xn--`, `sthree-`, `-s3alias`).

The result was that the same value meant different things to different operations.
Measured by building requests offline for the name `my_bucket`:

| Operation | Before | After |
| --- | --- | --- |
| `create_bucket` | accepted | rejected at the client method |
| `delete_bucket` | rejected at `send()` | rejected at the client method |
| `bucket_exists` | accepted | rejected at the client method |
| `put_bucket_versioning` | accepted | rejected at the client method |
| `get_bucket_versioning` | rejected at `send()` | rejected at the client method |
| `get_bucket_tagging` | accepted | rejected at the client method |
| `get_bucket_policy` | rejected at `send()` | rejected at the client method |
| `get_object`, `stat_object` | accepted | rejected at the client method |
| `delete_object` | rejected at `send()` | rejected at the client method |

`sthree-data`, `xn--bucket`, `My_Bucket` and `bucket:one` behave the same way.
Strict-valid names such as `my-bucket` and `my.bucket` were accepted everywhere
before and still are.

This change makes the constructor strict, which is what AIStor enforces, and
removes the 28 downstream checks that can no longer fire. `get_object_fast`
validated each argument twice, explicitly and again through the constructors; the
explicit pair is removed. `new_strict` delegates to `new` and keeps its meaning.
`check_bucket_name` keeps its `strict` parameter, since it is public API.

## Motivation and Context

A validated wrapper is worth its cost only when one rule set is applied in one
place. Two rule sets meant a caller could hold a `BucketName` and still not know
which operations would accept it, and the answer depended on whether a given
builder happened to carry the extra line.

Failing at `send()` rather than at the client method also puts the error far from
the input that caused it. The name is supplied to `client.delete_bucket(...)`, and
that call returned `Ok`.

## How to test this PR?
```
cargo fmt --all -- --check
cargo clippy --all-targets --workspace -- -D warnings
cargo test --lib
cargo test --doc
```
Two tests in `src/s3/types/typed_parameters.rs` guard the invariant, and both fail
before this change:

- `test_bucket_name_is_strict` rejects uppercase, `_`, `:` and the three reserved
  forms at construction.
- `test_bucket_name_constructors_agree` pins `new` and `new_strict` to one rule
  set, so the two cannot drift apart again.

Integration tests under `tests/s3/` need a live server. Their names come from
`rand_bucket_name()`, which produces `test-bucket-<uuid>` in lowercase, so they are
unaffected.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [x] Cleanup/Maintenance (no functional changes)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

One breaking change for callers:

| Item | Before | After |
| --- | --- | --- |
| `BucketName::new` | relaxed rules: uppercase, `_`, `:` and AWS-reserved names accepted | strict rules, same as `new_strict` |

Code that built such a name and used it with `get_object`, `create_bucket` or
another relaxed operation now gets `ValidationErr::InvalidBucketName` from the
client method. The same code already failed at `send()` for `delete_bucket`,
`delete_object`, `get_bucket_policy` and `get_bucket_versioning`.

## Checklist:

- [x] Fixes a regression (If yes, please add commit-id or PR # here) — c450081 (#200)
- [x] Unit tests added/updated
- [x] Internal documentation updated
- [ ] Create a documentation update request
- [x] No internode changes / version interoperability introduced

## Section for AI reading

Base 16feb86. Diff: 19 files, +63/-105.

The mismatch dates to #200, which replaced `bucket: String` with
`bucket: BucketName` in `builders/bucket_common.rs` and wrote `BucketName::new`
with `check_bucket_name(&name, false)`. The strict lines in the builders predate
it, going back to the builder pattern in #74, #76, #78 and #85, where the field
was a `String` and the check was the only validation. Those lines were correct
when written and became redundant in #200. #217 copied the pattern into the
inventory and QoS builders.

Removal was mechanical: every one of the 28 sites reads
`check_bucket_name(&self.bucket, true)?;` over a field typed `BucketName`, either
directly or through `BucketCommon`. 17 imports of `check_bucket_name` became
unused and are dropped. No call site signature changed.

`check_bucket_name(name, false)` now has no caller in `src/`. The relaxed regex
and the `strict` parameter are left in place because both are public API; removing
them belongs with the other breaking cleanups in docs/DESIGN-RULES-TODO.md.

Server responses are unaffected: they build names through
`BucketName::new_unchecked`, so a bucket created outside this SDK with a relaxed
name still parses.

Verified: cargo fmt --all -- --check clean; cargo clippy --all-targets clean, zero
warnings; cargo test --lib 554 passed, 0 failed; cargo test --doc 88 passed, 2
ignored. The rdma feature is Linux only, so --all-features is not usable on
Windows; this change does not touch it.

## Summary by CodeRabbit

* **Documentation**
  * Added design guidelines covering validation, typed errors, state handling, privacy, locking, and testing.
  * Added a backlog for tracking design-rule conformance work.

* **Improvements**
  * Bucket-name validation now rejects surrounding whitespace and additional reserved prefixes and suffixes.
  * Added validation and recognition for S3 Express directory bucket names.
  * Bucket names returned by the service are validated before being provided to callers.

* **Bug Fixes**
  * Invalid bucket names returned by the service now produce a validation error instead of being passed through unchecked.


HJLebbink:
@coderabbitai review

HJLebbink:
@coderabbitai full review

HJLebbink:
@coderabbitai what needs to be changed for you to approve this PR?

HJLebbink:
@coderabbitai full review

